PR #4, opened 2018-11-09
Fix use of isA

jamesr66a:
During build I was receiving warnings saying that the boolean value of the functions `isA`'s address will always evaluate true. I believe the intention here was to call `isA` rather than find the truthiness of its address


jspark1105:
Great catch @jamesr66a !

PR #42, opened 2018-12-06
Add missing <algorithm> include

jamesr66a:
std::max and std::min are defined in the standard library header <algorithm> (https://en.cppreference.com/w/cpp/algorithm/max), but this include is not present, leading to build failures on my mac.

### DIFF
--- a/src/FbgemmI8DepthwiseAvx2.cc
+++ b/src/FbgemmI8DepthwiseAvx2.cc
@@ -6,7 +6,7 @@
  */
 #include "FbgemmI8DepthwiseAvx2.h"
 
-#include <algorithm>
+#include <algorithm> // for min and max
 #include <cassert>
 #include <cmath> // for lrintf and sqrt
 #include <tuple> // for tie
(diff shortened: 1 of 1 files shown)


src/FbgemmI8DepthwiseAvx2.cc
```diff
@@ -6,6 +6,7 @@
  */
 #include "FbgemmI8DepthwiseAvx2.h"
 
+#include <algorithm>
```
dskhudia: Thanks for fixing this. We will soon have Mac builds in our CI so such issues get caught. Please add the following comment after the include. We want to keep track of stl headers in avx2/avx512 files and possibly remove them in future. We want to minimize the use of stl in files compiled with avx2/avx512 flags.
#include <algorithm> // for min and max

dskhudia: Sounds good.

jamesr66a: As of https://github.com/pytorch/pytorch/commit/65da7ddad65322b393fdd9549a349b3f7eb54049 we are now building FBGEMM by default in PyTorch. Everything is looking good on the build front

PR #44, opened 2018-12-06
Fix fb-config branch for FaceHub extension

yns88:
This branch is special and needs to contain just a browser-extension.json file in order for the FaceHub browser extension to show an "Import" button -- see pytorch/pytorch as an example: https://github.com/pytorch/pytorch/tree/fb-config


yns88:
Oh wait nevermind, the browser-extension.json file is there already...

PR #52, opened 2018-12-23
Fix a bug in FbgemmFP16

lifeiteng:
https://github.com/pytorch/FBGEMM/blob/master/test/FP16Test.cc#L46 will fail if batch_size (e.g. m) is larger than 120.


jspark1105:
@lifeiteng thanks for fixing this!

PR #56, opened 2019-01-08
FP16Benchmark: Allow fp32 comparison using cblas

WilliamTambellini:
FP16Benchmark: Allow comparison against fp32
using any local cblas library if MKL not found.


dskhudia:
Thanks William :) Let me merge it.

PR #82, opened 2019-03-01
Fixes for FBGEMM FP16 performance

jianyuh:
Summary:
This is a quick fix for matching FBGEMM FP16 performance with SKINNY GEMM FP16.

Basically, this Diff switches the register layout in C accumulation buffer inside micro-kernel from MR * 1 to MR * 2. Check the reasons in T40816746.

Differential Revision: D14278430


src/FbgemmFP16.cc
```diff
@@ -208,7 +200,7 @@ FBGEMM_API void cblas_gemm_compute(
   const int n = Bp.numCols(), k = Bp.numRows(), ldc = n;
   const int mb_max = 120;
   constexpr int simd_width = 8;
-  constexpr int kernel_ncol_blocks = 1;
+  constexpr int kernel_ncol_blocks = 2;
```
jspark1105: Maybe put this somewhere in FbgemmFP16.h and use it for when computing bcol_ ?

src/FbgemmFP16.cc
```diff
@@ -35,23 +35,15 @@ inline void PackA(int nrow, int ncol, const float* from, int ldim, float* to) {
...
-      gemmkernel_6x1_AVX2_fA0fB0fC0,
-      gemmkernel_7x1_AVX2_fA0fB0fC0,
-      gemmkernel_8x1_AVX2_fA0fB0fC0,
-      gemmkernel_9x1_AVX2_fA0fB0fC0,
-      gemmkernel_10x1_AVX2_fA0fB0fC0,
-      gemmkernel_11x1_AVX2_fA0fB0fC0,
-      gemmkernel_12x1_AVX2_fA0fB0fC0,
-      gemmkernel_13x1_AVX2_fA0fB0fC0,
-      gemmkernel_14x1_AVX2_fA0fB0fC0
+      gemmkernel_1x2_AVX2_fA0fB0fC0,
```
jspark1105: Add comment 2 in ?x2 should be same as kernel_ncol_blocks?

jianyuh: Added the comments.

src/FbgemmFP16.cc
jianyuh: Added "kernel_ncol_blocks" in FbgemmFP16.h and used it for computing bcol_.

src/FbgemmFP16.cc
```diff
@@ -291,35 +283,34 @@ FBGEMM_API void cblas_gemm_compute(
...
+              if ((rem % Bp.blockColSize()) == 0) {
                 gp.B = &(Bp(k_ind, last_blk_col));
                 gp.C = &C[m2 * ldc + last_blk_col];
                 gp.b_block_cols = 1;
                 KernelInfo::kernel[kernel_nrows](&gp);
               } else {
                 // small temporary buffer
-                float c_tmp[16 * 24] = {0};
-                assert((16 * 24) > kernel_nrows * kernel_ncols);
+                float c_tmp[16 * 2 * 16 * 6] = {0};
```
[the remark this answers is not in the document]
jianyuh: Sorry I made it too large. The size of the temporary buffer size should be larger than the required kernel_nrow x kernel_ncols elements computed in the registers. The original 16 * 24 works for the new kernels, as currently the maximum kernel_nrow = 6 and  the kernel_ncols = 16.

jianyuh:
> Can we also update codegen_fp16fp32.cc ?
Updated "codegen_fp16fp32.cc".

PR #93, opened 2019-04-15
group conv using intrinsic

jspark1105:
Summary:
The current JIT group conv implementation was a bit inflexible to experiment with various loop tiling schemes.
This diff replaces the JIT implementation with intrinsics, which will also facilitate using AVX512.
Once we know what's a good loop tiling scheme, packing order for weight matrix, and so on, we may want to try JIT again.

Here's the loop order we're using

for each MB // i0
  for each G
   for each H // i1
    // After each iteration, we run output processing. We could do this for each K loop iteration, but we run output processing for every 8 groups for spatial locality and vectorization
    for each g1 // g1 block size is 8
     for each K // j0
      for each C // k0
       // Load gRSck weights into registers reused by W times
       for each W // i2
        // for each iteration at this level, we have a complete row offset
        for each RS // j1
         // Activation is broadcasted into registers and reused by k times
         // Output is accumulated in-register by RSc times
         --- Below is vectorization
         for each g2 // block size 2 if C_per_G == 4 or 1
          for each k // j2, block size 8
           for each c // k2, block size 4

Differential Revision: D14932309


src/GroupwiseConv.cc
```diff
@@ -0,0 +1,841 @@
...
+#endif
+
+  static_assert(SPATIAL_DIM == 2, "3D conv not supported yet");
+
+  assert(G % 8 == 0);
+  // Overall loop order
+  // for each MB // i0
+  //  for each G
+  //   for each H // i1
+  //    for each g1
```
protonu: what is g1 and g2?
what is g in gRSck?

jspark1105: Sorry I was sloppy when I was writing this comment. Will update soon. Upper case is outer loop for that dimension and lower case is inner loops for that dimension. For group dimension, we have 3 levels of loop, G, g1, and g2. g2 is for vectorization (only used when we don't have enough channels per group to fill the whole SIMD register), and g1 is for vectorizing the computation of row_offset (we need to work on 8 groups at a time if we vectorize row_offset computation across groups).
R = vertical filter size (usually 1 or 3)
S = horizontal filter size (usually 1 or 3)

Issue #87, opened 2019-03-18
use git submodules


t-vi:
Would it be possible to use git submodules for ASMJIT?
The downloading during the (PyTorch) build can be a bit hard for me when hacking PyTorch while travelling.
Thanks!

dskhudia:
I think so. For now, you can download asmjit once and point ASMJIT_SRC_DIR env var to the downloaded source code to avoid downloading with build. Particular git tag we use is in https://github.com/pytorch/FBGEMM/blob/master/cmake/modules/DownloadASMJIT.cmake

dhpollack:
@t-vi I run into this problem all the time too.  I created a pull request.  Perhaps you could test it out and let me know if it works for you.

dskhudia:
@dhpollack Thanks a lot for the pull request. Let me test its impact on Caffe2/Pytorch build.

dskhudia:
Update: I had to revert the commit because adding submodules involves one more step so that internal tools work correctly. Will re-commit soon.

dhpollack:
@dskhudia I also a few other things that I can update before you commit this pr.

1. I may want to use `${FBGEMM_SOURCE_DIR}` instead of `${CMAKE_SOURCE_DIR}`, because I'm not sure what the latter is when FBGEMM is a submodule itself.
2. You may want to consider renaming `${ASMJIT_SRC_DIR}` to `${ASMJIT_SOURCE_DIR}` to make it consistent with the other variables.

Let me know if you want me to make those changes.  I am not sure it matters as both work for me right now.

dskhudia:
1) I will change it to FBGEMM_SOURCE_DIR
2) I remember it being having an issue with ASMJIT's cmake file. So we should keep that.

dskhudia:
@t-vi and @dhpollack : It's fixed now. Once the pytorch's fbgemm submodule gets updated (should happen soon) to fbgemm's latest commit, pytorch build will not download asmjit at build time. 

@dhpollack : Thanks for your PR 👍

PR #95, opened 2019-05-10
switch from cmake downloads to git submodules

dhpollack:
I created a pull request for #87.  I also tend to do a lot of hacking without an internet connection and it is nice to have the required library offline.  I also get a cryptic error message when I build pytorch without an internet connection because these modules aren't available.


dskhudia:
This looks good to me. I'll merge it soon.

PR #107, opened 2019-07-13
Quantize should work even if the value after quantization becomes inf

dskhudia:
Summary:
With small `scale` and large `src` (value to be quantized) the `src/scale` may become +inf or -inf. We should quantized to max/min in this case.

Without this fix, we get the following error.

```
> buck-out/dev/gen/deeplearning/fbgemm/fbgemm_generic#header-mode-symlink-tree-with-header-map,headers/fbgemm/QuantUtils.h:66:35: runtime error: inf is outside the range of representable values of type 'long'

```

Differential Revision: D16234008

### DIFF
--- a/include/fbgemm/QuantUtils.h
+++ b/include/fbgemm/QuantUtils.h
@@ -56,10 +56,8 @@ T Quantize(
     int result_precision,
     bool result_is_signed = std::is_signed<T>::value) {
   const float transformed_val = zero_point + src / scale;
-  return clamp<std::int64_t, T>(
-      static_cast<std::int64_t>(std::nearbyint(transformed_val)),
-      result_precision,
-      result_is_signed);
+  return clamp<float, T>(
+      std::nearbyint(transformed_val), result_precision, result_is_signed);
 }
 
 template <typename T>
--- a/test/QuantUtilsTest.cc
+++ b/test/QuantUtilsTest.cc
@@ -272,3 +272,38 @@ TEST(QuantizeTestSingle, vectorScalar) {
   EXPECT_TRUE(
       adjacent_find(dst.begin(), dst.end(), not_equal_to<int>()) == dst.end());
 }
+
+TEST(QuantizeTest, cornerCases) {
+  TensorQuantizationParams qparams;
+  qparams.scale = 1.19209e-07;
+  qparams.zero_point = 0;
+  qparams.precision = 8;
+  std::vector<float> src1 = {3.40282e+38, -2.16845e+38};
+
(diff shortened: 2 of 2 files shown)


jspark1105:
I feel it can be somewhat dangerous to kind of silence overflow in some cases. This will also introduce inconsistencies with optimized vectorized code. Maybe it's better to ask caller to make sure there's no overflow? Did you see inf or -inf in real use cases or just from stress tests?

include/fbgemm/QuantUtils.h
```diff
@@ -60,7 +60,12 @@ FBGEMM_API T Quantize(
     float scale,
     int result_precision,
     bool result_is_signed = std::is_signed<T>::value) {
-  const float transformed_val = zero_point + src / scale;
+  float transformed_val = std::nearbyintf(zero_point + src / scale);
```
jerryzh168: @zafartahirov and I addressed this problem in: https://github.com/pytorch/pytorch/pull/22830

jerryzh168: we are making sure no overflow in caller site

dskhudia:
> I feel it can be somewhat dangerous to kind of silence overflow in some cases. This will also introduce inconsistencies with optimized vectorized code. Maybe it's better to ask caller to make sure there's no overflow? Did you see inf or -inf in real use cases or just from stress tests?

I only observed in the stress tests. And Jerry mentioned that it's been fixed at the call site. I'll abandon this diff.

test/QuantUtilsTest.cc
```diff
@@ -181,3 +181,32 @@ TEST_P(QuantizeGroupwiseTest, quantizeTest) {
...
+  qparams.precision = 8;
+  std::vector<float> src1 = {3.40282e+38, -2.16845e+38};
+
+  std::vector<int8_t> dst_int8(src1.size());
+  Quantize<int8_t>(src1.data(), dst_int8.data(), dst_int8.size(), qparams);
+  EXPECT_EQ(dst_int8[0], 127);
+  EXPECT_EQ(dst_int8[1], -128);
+
+  // Tests vectorized and remainder paths
+  std::vector<float> src2 = {3.40282e+38,
```
zafartahirov: These are infinities? If the test is for the corner cases, it should cover other tests cases as well (s.a. extremely small, extremely small negative, etc.)

dskhudia: I have postponed this diff for now. It needs more discussion on numeric mismatch between scalar and vector code.

PR #108, opened 2019-07-17
Support pointwise with unified convolution interface as well

dskhudia:
Summary: Pointwise gets converted to direct GEMM

Differential Revision: D16296356


jianyuh:
LGTM! It appears that we don't have "approve" button for FBGEMM?

PR #120, opened 2019-08-29
int8 specialization for AVX2 Quantize routine

jamesr66a:
This adds a specialization for `int8` to the AVX2 `Quantize` routine.

I tried also adding a specialization for `int32` (the final datatype we support in PyTorch quantization), but it seemed to introduce numerical issues stemming from the difference in implementations:

https://github.com/pytorch/FBGEMM/blob/master/include/fbgemm/QuantUtils.h#L63

vs

https://github.com/pytorch/FBGEMM/blob/master/src/QuantUtilsAvx2.cc#L82

### DIFF
--- a/include/fbgemm/QuantUtilsAvx2.h
+++ b/include/fbgemm/QuantUtilsAvx2.h
@@ -40,9 +40,10 @@ struct FBGEMM_API RequantizationParams {
 ////////////////////////////////////////////////////////////////////////////////
 // Utility functions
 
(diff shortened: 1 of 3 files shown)


dskhudia:
Let me take a look at the int32 case. I have one diff related to this: https://github.com/pytorch/FBGEMM/pull/107/files The overall goal is keep the numerics same between vectorized and scalar code.

PR #151, opened 2019-10-26
moving SpMM kernels from experimental to fbgemm

jspark1105:
Differential Revision: D18137995


src/FbgemmSpMM.cc
```diff
@@ -0,0 +1,721 @@
...
+
+namespace x86 = asmjit::x86;
+
+using internal::SpMMTypeTrait;
+
+namespace {
+
+template <class To, class From>
+typename enable_if<
+    (sizeof(To) == sizeof(From)) && is_trivially_copyable<From>::value &&
```
jamesr66a: FYI `std::is_trivially_copyable` isn't supported by gcc-4's stdlib. I mention this because in README it says FBGEMM needs GCC 4.9+

dskhudia: Is this causing a build failure for PyTorch is some configuration?

jamesr66a: No, PyTorch has dropped gcc-4 support, so this isn't breaking PT

PR #153, opened 2019-10-28
Add support for AVX512_256 into fbgemm fp16

jspark1105:
Differential Revision: D18138146


include/fbgemm/Utils.h
```diff
@@ -15,6 +15,17 @@
 
 namespace fbgemm {
 
+/**
+ * @brief Helper typedef for supported x86 ISA.
+ */
+enum ISA {
```
jspark1105: Can we reuse inst_set_t in line 46?

src/FbgemmFP16UKernelsAvx2.cc
```diff
@@ -8,7 +8,7 @@
 
 namespace fbgemm {
 
-void __attribute__((noinline)) gemmkernel_1x2_AVX2_fA0fB0fC0(GemmParams* gp) {
+void __attribute__((noinline)) gemmkernel_1x2_Avx2_fA0fB0fC0(GemmParams* gp) {
```
jspark1105: Is it possible to have a separate commit that just do cosmetic changes like AVX2 -> Avx2? This will help us track changes later.

include/fbgemm/Utils.h
```diff
@@ -15,6 +15,17 @@
...
+  unknown = -1,
+  avx2 = 0,
+  avx512 = 1,
+  avx512_256 = 3,
```
jspark1105: Can we add a comment on what avx512_256 is?

src/FbgemmFP16.cc
```diff
@@ -410,10 +437,9 @@ FBGEMM_API void cblas_gemm_compute(
 
   GemmParams gp;
 
-  const funcptr_fp16* kernels =
-      use_avx512 ? KernelInfo::kernel_avx512 : KernelInfo::kernel_avx2;
-  const array<array<array<int, 2>, 2>, 121>& partition =
-      use_avx512 ? KernelInfo::partition_avx512 : KernelInfo::partition_avx2;
+  auto const isaIndex = simd_width == 16 || iset == avx2 ? iset : avx512_256;
```
jspark1105: Why can't we just use iset directly?

include/fbgemm/Utils.h
efiks: Sure, will do.

src/FbgemmFP16UKernelsAvx2.cc
efiks: Sure, will do

src/FbgemmFP16.cc
efiks: yes

PR #169, opened 2019-11-08
Add include/fbgemm/FbgemmI8DepthwiseAvx2.h to public library

zrphercule:
When using fbgemm externally we need to include include/fbgemm/FbgemmI8DepthwiseAvx2.h as well, however it is not public now.


dskhudia:
Thanks for fixing.

PR #170, opened 2019-11-10
Enable the n dimensional parallelization

jianyuh:
Summary:
Enabling the parallelization on the n dimension in FBGEMM. This is useful when we have the multi-thread Predictor and the target matrix shapes have large n dimension sizes. With this Diff, we can also parallelize both m and n dimensions (2D parallelization) simultaneously.

When parallelizing the n dimension, note that we have the additional overhead of allocating the duplicated packA buffer for each thread.

Differential Revision: D17696021

### DIFF
--- a/include/fbgemm/Utils.h
+++ b/include/fbgemm/Utils.h
@@ -162,6 +162,58 @@ struct FBGEMM_API BlockingFactors {
   int NCB;
 };
 
+/**
+ * @brief A struct to represent the partition information for the threads on the
+ * m and n dimensions.
+ */
+struct FBGEMM_API thread_type_t {
+  int g_num_threads;
+  int m_num_threads;
+  int n_num_threads;
+  int g_thread_id;
+  int m_thread_id;
+  int n_thread_id;
+
+  std::string toString() const {
+    std::string out = "";
+    out += "g num threads: " + std::to_string(g_num_threads) + ", ";
+    out += "m num threads: " + std::to_string(m_num_threads) + ", ";
+    out += "n num threads: " + std::to_string(n_num_threads) + ", ";
+    out += "g thread id: " + std::to_string(g_thread_id) + ", ";
+    out += "m thread id: " + std::to_string(m_thread_id) + ", ";
+    out += "n thread id: " + std::to_string(n_thread_id);
+    return out;
+  }
+};
+
+/**
+ * @brief A heuristic algorithm to partition the threads across m and n
+ * dimensions for parallelization, ensuring the ratio between the number of rows
+ * allocated to each thread in the m dimension and the number of columns
+ * allocated to each thread in the n dimension is approximately aspect_ratio.
+ *
+ * The less aspect_ratio is, the more favorable it is to parallelize the m
+ * dimension over the n dimension.
+ */
+FBGEMM_API int fbgemmGet2DPartition(
+    int m,
+    int n,
+    int nthreads,
+    int n_align,
+    double aspect_ratio);
+
+/**
+ * @brief A heuristic way to partition the threads across g, m and n dimensions
+ * for parallelization.
+ */
+FBGEMM_API thread_type_t fbgemmGetThreadPartition(
+    int g,
+    int m,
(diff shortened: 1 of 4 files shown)


src/Fbgemm.cc
```diff
@@ -136,28 +137,42 @@ void fbgemmPacked(
   t_very_start = std::chrono::high_resolution_clock::now();
 #endif
 
+  thread_type_t thread_partition_info =
+      fbgemmGetThreadPartition(MDim, NDim, thread_id, num_threads);
```
jspark1105: We first do 2D partition assuming G == 1 and then partition across m when there're multiple groups after the fact. We may want to first decide how many threads will be used for each group and then 2D-partition based on that? In this way, we should be able to do globally better partitioning.

src/Utils.cc
```diff
@@ -252,4 +252,51 @@ void* fbgemmAlignedAlloc(
...
+  if (m < n && m / num_threads < 32 &&
+      n / num_threads >=
+          128) { // Only parallelize the n dimension. Note that we have the
+                 // additional overhead of allocating the duplicated packA
+                 // buffer for each thread when parallelizing this dimension.
+    m_num_threads = 1;
+  } else if (
+      num_threads >= 4 && m / num_threads < 128 &&
+      n / num_threads >= 64) { // Parallelize both m and n dimension.
+    m_num_threads = static_cast<int>(std::sqrt(num_threads));
```
jspark1105: We should prefer parallelizing across m so use ceil?

src/Utils.cc
```diff
@@ -252,4 +252,51 @@ void* fbgemmAlignedAlloc(
...
 
+thread_type_t
+fbgemmGetThreadPartition(int m, int n, int thread_id, int num_threads) {
+  int m_num_threads;
+  assert(num_threads >= 1);
+  // Heuristic for determine the thread partitions for parallelizing across m or
+  // n dimensions.
+  // TODO: more smart ways for thread partitions considering the
+  // grain size (MR, NR) parameters
+  if (m < n && m / num_threads < 32 &&
```
jspark1105: Why do we need this special case? Even when this if condition is met, we still want to parallelize across m when possible as long as m assigned for each thread is not much smaller than n assigned for each thread?

include/fbgemm/Utils.h
```diff
@@ -162,6 +162,33 @@ struct FBGEMM_API BlockingFactors {
...
+    return out;
+  }
+};
+
+/**
+ * @brief A heuristic way to partition the threads across m and n dimensions for
+ * parallelization.
+ */
+FBGEMM_API thread_type_t
+fbgemmGetThreadPartition(int m, int n, int num_threads, int thread_id);
```
jspark1105: We may want to pass n_align argument (default value to be 64) so each n partition is cache line aligned.

src/Fbgemm.cc
jianyuh: Created a "3D" thread partition function to select all the thread partition number across g, m, and n dimensions.

src/Utils.cc
jianyuh: Updated.

include/fbgemm/Utils.h
jianyuh: Added n_align as one additional parameter. When we also partition along the n dimension and align the cache line (64), we need to do more careful design for the thread partition heuristics. Didn't change the code yet...

src/Utils.cc
jianyuh: This case is for purely parallelizing the n dimension, which is needed for the language models (where m << n, k).

jianyuh: Update the heuristic algorithms for allocating the number of threads along the m and n dimensions, and removed this special case

include/fbgemm/Utils.h
jianyuh: Changed the code and added the handler for the alignment of cache line and avoid the false sharing due to the wrong cache line border case.

PR #203, opened 2019-12-02
Migrate the FP16/BF16 <-> FP32 conversion routines into OSS

jianyuh:
Summary: As Title says, we would like to move FP16/BF16 <-> FP32 conversion routines into OSS.

Differential Revision: D18766898


jianyuh:
We added the PyTorch CI test PR: https://github.com/pytorch/pytorch/pull/30634

PR #237, opened 2020-01-08
Githubaction

shz0116:
Fix the cmake for OpenMP setting.
Fix some link errors for undefined references.
Disable running actions on macos-latest now.


src/Fbgemm.cc
```diff
@@ -267,7 +267,7 @@ bool fbgemmSupportedCPU() {
 ////////////////////////////////////////////////////////////////////////////////
 // ReQuantizeOutput
 #define INSTANTIATE_BASE(PACK_A, ACC_T, RELU, Q_GRAN, BIAS_TYPE)    \
-  template void fbgemmPacked(                                       \
+  template FBGEMM_API void fbgemmPacked(                                       \
```
dskhudia: @shz0116 We only need to add FBGEMM_API in the header file. Please add in header file and remove all these from here.

shz0116: Have you tested ? They are needed here.  DGEMM_API has already been defined in header file before.
This is tested on Mac, not sure how on Linux.

PR #238, opened 2020-01-08
[WIP] Enable MSVC build

### LINKED ISSUE #150: Is this available on windows?
Can this library be built on windows?

peterjc123:
Fixes #150.

Worklist:

- [x] Fix CMake script
- [x] Remove wrong annotations of `FBGEMM_API`
- [x] Replace VLAs with `_aligned_malloc` and `_aligned_free`
- [ ] Write assembly code for Windows (I don't think I'm capable of doing that.)

### DIFF
--- a/include/fbgemm/QuantUtils.h
+++ b/include/fbgemm/QuantUtils.h
@@ -32,7 +32,7 @@ FBGEMM_API void ChooseRequantizationMultiplier(
 // TODO: T26263653 fix signed-integer-overflow undefined behavior
 template <typename T1, typename T2 = std::uint8_t>
 NO_SANITIZE("signed-integer-overflow")
-FBGEMM_API T2 clamp(T1 src, int precision, bool is_signed = false) {
+T2 clamp(T1 src, int precision, bool is_signed = false) {
   std::int32_t min = is_signed ? -(1LL << (precision - 1)) : 0;
   std::int32_t max =
       is_signed ? ((1LL << (precision - 1)) - 1) : (1LL << precision) - 1;
@@ -49,7 +49,7 @@ FBGEMM_API T2 clamp(T1 src, int precision, bool is_signed = false) {
 /// Quantize src using zero_point and scale, clamp to the specified precision,
 /// and convert it to type T
 template <typename T>
-FBGEMM_API T Quantize(
+T Quantize(
     float src,
     std::int32_t zero_point,
     float scale,
@@ -63,7 +63,7 @@ FBGEMM_API T Quantize(
 }
 
 template <typename T>
-FBGEMM_API T Quantize(float src, const TensorQuantizationParams& qparams) {
+T Quantize(float src, const TensorQuantizationParams& qparams) {
   return Quantize<T>(src, qparams.zero_point, qparams.scale, qparams.precision);
 }
 
@@ -109,12 +109,12 @@ FBGEMM_API void QuantizeGroupwise(
     T* dst);
 
 template <typename T>
-FBGEMM_API float Dequantize(T src, const TensorQuantizationParams& qparams) {
+float Dequantize(T src, const TensorQuantizationParams& qparams) {
   return qparams.scale * (src - qparams.zero_point);
 }
 
 template <typename T>
-FBGEMM_API void Dequantize(
+void Dequantize(
     const T* src,
     float* dst,
     int len,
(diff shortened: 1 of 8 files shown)


peterjc123:
Current build errors:
```
src/FbgemmFP16UKernelsAvx2.cc(443,7): error C2059: syntax error: 'volatile'
src/FbgemmFP16UKernelsAvx2.cc(625,7): error C2059: syntax error: 'volatile'
... and more
```
Explanation:
The gcc style of assembly is very different from the msvc style one. It needs a complete rewrite. BTW, inline assembly is not supported in an x64 project for msvc, too.

src/PackDepthwiseConvMatrixAvx2.cc
```diff
@@ -88,7 +89,13 @@ PackedDepthWiseConvMatrix::PackedDepthWiseConvMatrix(
   // (12, 8), (12, 9), (12, 10), zero, ..., (15, 8), (15, 9), (15, 10), zero
   // (28, 8), (28, 9), (28, 10), zero, ..., (31, 8), (31, 9), (31, 10), zero
   for (int k1 = 0; k1 < K; k1 += 32) {
+    #ifdef _MSC_VER
```
dskhudia: fbgemmAlignedAlloc and fbgemmAlignedFree should take care of it

include/fbgemm/QuantUtils.h
```diff
@@ -131,7 +131,7 @@ FBGEMM_API std::int64_t
 SaturatingRoundingMulWithShift(std::int32_t a, std::int32_t b, int right_shift);
 
 template <typename T>
-FBGEMM_API T Requantize(
+T Requantize(
```
dskhudia: These are needed for shared library build on Linux. What problems do they create on windows?

dskhudia:
> The gcc style of assembly is very different from the msvc style one. It needs a complete rewrite. BTW, inline assembly is not supported in an x64 project for msvc, too.

The current plan is to not have assembly version for windows. We will add a C/C++ only inefficient version for now. Eventually the inline assembly will go away for gcc as well as MSVC as we move to runtime code generation for fp16 GEMMs.

ykim362:
> The gcc style of assembly is very different from the msvc style one. It needs a complete rewrite. BTW, inline assembly is not supported in an x64 project for msvc, too.

I have another solution which replaces each asm line to intrinsic APIs line by line. 
e.g. https://github.com/marian-nmt/FBGEMM/blob/master/src/codegen_fp16fp32.cc#L284
I know this is not the best solution and jit might be a better solution. But, this works just fine without losing performance on both Linux and windows.

KaiyuYue:
> > The gcc style of assembly is very different from the msvc style one. It needs a complete rewrite. BTW, inline assembly is not supported in an x64 project for msvc, too.
> 
> I have another solution which replaces each asm line to intrinsic APIs line by line.
> e.g. https://github.com/marian-nmt/FBGEMM/blob/master/src/codegen_fp16fp32.cc#L284
> https://github.com/marian-nmt/FBGEMM/blob/master/src/FbgemmFP16UKernelsAvx2.cc#L125
> 
> I know this is not the best solution and jit might be a better solution, eventually. But, this works just fine without losing performance on both Linux and windows before the jit code coming.

Could I compile the fbgemm with your forked version successfully on windows?

include/fbgemm/QuantUtils.h
peterjc123: They are not needed because:
1. They are the in the header file so they will be visible at the user side.
2. They have the implementation, not just declarations.

peterjc123:
@ykim362 So are you interested in creating a new PR or getting the essential changes in this PR?

ykim362:
@peterjc123 I've been working on AVX512 fp16 kernels. I will submit a PR soon. Thanks for bringing this up.

dskhudia:
@peterjc123 Thanks for your contributions. I have incorporated your changes into the following PRs with minor modifications. These PRs also take care of inline assembly by excluding it for MSVC and redirecting the kernel calls to slower reference implementation. 
 
https://github.com/pytorch/FBGEMM/pull/240
https://github.com/pytorch/FBGEMM/pull/241
https://github.com/pytorch/FBGEMM/pull/242

The PRs take care of the build for the core library. We still need to make sure it runs fine as well. 
@ykim362 Thank you for making it work on windows in your fork. I will be picking up some changes from your fork to make sure the runtime is also fine.

peterjc123:
@dskhudia Great work! What about introducing a Windows CI with the build-only job that shows the current status of the work towards Windows?

dskhudia:
Thanks. @shz0116 is working on adding a windows build-only job using github Actions. If we face issues with setting it up with github Actions, we will appreciate your help in doing it using CircleCI.

PR #244, opened 2020-01-11
Add Threading Support for QuanUtils

shz0116:
Add Threading for QuanUtils


include/fbgemm/FbgemmBuild.h
```diff
@@ -68,3 +68,11 @@
 #if !defined(NO_SANITIZE)
 #define NO_SANITIZE(what)
 #endif
+
+// handle special export requirements for shared build on Macos
+#if __clang__
+#define FBGEMM_API_MACOS __attribute__((__visibility__("default")))
```
dskhudia: We shouldn't create another one of  these. It will only cause more confusion.

shz0116:
Implemented in other PRs.

PR #247, opened 2020-01-13
Fix undefined reference error on macos

shz0116:
Fix the reference undefined error on macos caused by enum class in Utils.h and UtilsAvx2.h


include/fbgemm/FbgemmBuild.h
```diff
@@ -26,7 +26,7 @@
       #endif
     #endif
   #else
-    #if __clang__ || __GNUC__ >=4 || __INTEL_COMPILER
+    #if __clang__ || __GNUC__ >=6 || __INTEL_COMPILER
```
dskhudia: Please add a comment here why we need GNUC > 6.

dskhudia:
@shz0116 Other than the comment, it looks good. Please start a run of PyTorch OSS tests as discussed.

include/fbgemm/FbgemmBuild.h
shz0116: added

PR #253, opened 2020-01-18
Add github action for windows

shz0116:


.github/workflows/fbgemmci.yml
```diff
@@ -68,3 +68,40 @@ jobs:
...
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
+        echo "INSTALL NINJA:"
+        pip install ninja
+        which ninja
+        mkdir build_static
+        cd build_static
+        echo "STARTING CMAKE"
+        cmake -G Ninja -DFBGEMM_BUILD_BENCHMARKS=OFF -DFBGEMM_BUILD_TESTS=OFF -DFBGEMM_LIBRARY_TYPE=static -DCMAKE_C_COMPILER="cl.exe" -DCMAKE_CXX_COMPILER="cl.exe" ..
+        ninja all
+        echo "Build Success"
```
dskhudia: Does it only echo Build Success if the previous command succeeds?

shz0116: I think so since whenever error happens, ninja will stop the whole process.

.github/workflows/fbgemmci.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         make
 
     - name: Test static FBGEMM lib
-      if: contains(runner.os, 'Linux')   # not run on macos-latest now due to supporting AVX2
+      if: contains(runner.os, 'linux')   # not run on macos-latest now due to supporting AVX2
```
jspark1105: contains(runner.os, 'Linux') doesn't work?

.github/workflows/fbgemmci.yml
```diff
@@ -37,6 +36,7 @@ jobs:
         echo GITHUB_WORKSPACE  = $GITHUB_WORKSPACE
         echo GITHUB_SHA        = $GITHUB_SHA
         echo GITHUB_REF        = $GITHUB_REF
+        c++ --verbose
```
jspark1105: Just for my education (I'm new to github action). A reason for adding this line?

include/fbgemm/FbgemmI8DepthwiseAvx2.h
```diff
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <cstdint>
+#include <stdexcept>  // needed for Microsoft VC 2019 Enterprise version on windows-latest
```
jspark1105: Actually I don't see any dependency to stdexcept in this file. Can you share compiler error message? Maybe the dependency is actually in one of the other header/source files including this. We want to minimize header includes from platform dependent files (*Avx2.{cc,h} or *Avx512.{cc,h}).

.github/workflows/fbgemmci.yml
shz0116: It works. The purpose here is to run Test only on Linux platforms now, not macos. Macos-latest does not support AVX2.

.github/workflows/fbgemmci.yml
shz0116: Just wan to print out some information. We can remove this later.

include/fbgemm/FbgemmI8DepthwiseAvx2.h
shz0116: Here is the error. I created a paste.
https://our.internmc.facebook.com/intern/paste/P125233423/

jspark1105: Got it. In these case, please include <stdexcept> from FbgemmI8Depthwise3x3Avx2.cc (and other files if needed). We want to minimize the number of headers included from a header file, unless code in header file actually uses the header file.

.github/workflows/fbgemmci.yml
jspark1105: Yes I understand the purpose of this line. I was just curious about why you changed Linux to linux

shz0116: It's not case sensitive. Just to be consistent with others.

include/fbgemm/FbgemmI8DepthwiseAvx2.h
shz0116: I moved stdexcept.h to cc files.

src/FbgemmI8Depthwise3DAvx2.cc
```diff
@@ -9,6 +9,7 @@
 
 #include <string>
 #include <tuple> // for tie
+#include <stdexcept> // for logic_error
```
jspark1105: nit: can you move stdexcept before string? In general we should run clang formatter before checking in the code. In VSCode you can easily run the formatter.

.github/workflows/fbgemmci.yml
peterjc123: Wait, isn't this cmd? For commands running in CMD, it won't stop on error. You may have to check manually with something like `if errorlevel 1 exit /b 1`.

shz0116: I will made the changes. Thx.

PR #254, opened 2020-01-18
Intrinsic implementation of FP16 kernels for windows build

ykim362:
This PR includes instrinsic API based implementation of FP16 kernels for windows build.
AVX2 is slightly slower or similar speed to the inline assembly version.
AVX512 is slower than inline assembly version for large m size.

Anyway, they are much faster than the reference implementation.

I've done accuracy tests for AVX2 and AVX512.

This is for #150 .


dskhudia:
Hi @ykim362, Thanks for the intrinsic version. I was wondering if you modified codegen? https://github.com/pytorch/FBGEMM/blob/master/src/codegen_fp16fp32.cc generates the code in the assembly files and we usually don't modify these files manually as future changes in codegen will result in more manual merge.

ykim362:
@dskhudia I directly changed the cc files, but I can add those changes to the codegen file.

dskhudia:
sounds good. Right now if you build codegen_fp16fp32.cc with g++ --std=c++11 fbgemm/src/codegen_fp16fp32.cc -o code_gen and run with ./code_gen it generates exactly the same code in src/FbgemmFP16UKernelsAvx2.cc, src/FbgemmFP16UKernelsAvx512.cc, src/FbgemmFP16UKernelsAvx512_256.cc and headers. We should have a similar behavior.

include/fbgemm/FbgemmBuild.h
```diff
@@ -55,7 +55,7 @@
 #if __clang__ || __GNUC__ >= 4 || __INTEL_COMPILER
 #define ALWAYS_INLINE inline __attribute__((__always_inline__))
 #elif _MSC_VER
-#define ALWAYS_INLINE __forceinline
+#define ALWAYS_INLINE // __forceinline
```
jspark1105: Can you tell me why we don't want to force inline in MSVC? Code size? Or some compile errors?

src/FbgemmFP16UKernelsAvx2.cc
```diff
@@ -1041,5 +1045,94 @@ gemmkernel_6x2_Avx2_fp16_fA0fB0fC0(GemmParamsFP16* gp) {
...
+      idxB += 16;
+
+      // first element
+      if (kk == 0) {
+        if(gp->beta != 0) {  // accumulate
+          for(size_t jj = 0; jj < kernel_nrows; jj++) {
+            // load A
+            __m256 ymmA = _mm256_broadcastss_ps(_mm_broadcast_ss((float const*)(gp->A + idxA + jj)));
+            // C = A * B + beta * C
+            ymmSum[2 * jj] = _mm256_fmadd_ps(ymmA, ymmB0, _mm256_mul_ps(ymmBeta, _mm256_loadu_ps(gp->C + idxC + jj * ldc_floatsize)));
```
jspark1105: If it's not too much trouble, can we move this functions to a separate file because this file is generated by a Python script? Also doing clang formatter will be really appreciated.

ykim362:
@jspark1105, @dskhudia
Thanks for open sourcing and maintaining this great library!

I agree with you for the approach to split files. Then, we might not need to put intrinsic codes into the codegen.
And, please feel free to copy and make another PR. (It would be great if the commit history remains.)

Thanks!

include/fbgemm/FbgemmBuild.h
ykim362: This causes infinite compilation time on MSVC at least for me. I am wondering if other people see the same problem.

PR #259, opened 2020-01-18
modified instrinsic fp16 kernel for windows build

jspark1105:
Summary: As title

Differential Revision: D19460978


ykim362:
The intrinsics based implementation files are missing.
Otherwise, it looks good.
Thanks!

PR #271, opened 2020-01-28
Add instantiation for template parameters

shz0116:


src/Fbgemm.cc
```diff
@@ -255,6 +255,54 @@ bool fbgemmSupportedCPU() {
...
+  template class FBGEMM_API ReQuantizeOutput<RELU, Q_GRAN, BIAS_TYPE>;
+
+#define INSTANTIATE_BIAS_T(RELU, Q_GRAN)       \
+  INSTANTIATE_BASE(RELU, Q_GRAN, std::int32_t) \
+  INSTANTIATE_BASE(RELU, Q_GRAN, float)
+
+#define INSTANTIATE_Q_GRAN(RELU)   \
+  INSTANTIATE_BIAS_T(RELU, QuantizationGranularity::TENSOR);      \
+  INSTANTIATE_BIAS_T(RELU, QuantizationGranularity::GROUP);       \
+  INSTANTIATE_BIAS_T(RELU, QuantizationGranularity::OUT_CHANNEL); \
```
jianyuh: Remove the trailing "\\".

.github/workflows/fbgemmci.yml
```diff
@@ -61,8 +61,8 @@ jobs:
         cmake -DFBGEMM_LIBRARY_TYPE=shared ..
         make
 
-    - name: Test shared FBGEMM  lib
-      if: contains(runner.os, 'Linux')   # not run on macos-latest now due to supporting AVX2
+    - name: Test shared FBGEMM lib
+      if: contains(runner.os, 'linux')   # not run on macos-latest now due to supporting AVX2
```
jianyuh: For the comment: does it mean "we cannot run on MacOS-latest, because it doesn't have support for AVX2"?

src/Fbgemm.cc
```diff
@@ -255,6 +255,54 @@ bool fbgemmSupportedCPU() {
...
+#undef INSTANTIATE_BASE
+
+#define INSTANTIATE_BASE(RELU, Q_GRAN, BIAS_TYPE)    \
+  template class FBGEMM_API ReQuantizeOutput<RELU, Q_GRAN, BIAS_TYPE>;
+
+#define INSTANTIATE_BIAS_T(RELU, Q_GRAN)       \
+  INSTANTIATE_BASE(RELU, Q_GRAN, std::int32_t) \
+  INSTANTIATE_BASE(RELU, Q_GRAN, float)
+
+#define INSTANTIATE_Q_GRAN(RELU)   \
```
jianyuh: Use "clang-format -i *.cc *.h" to apply the clang format here.

src/Fbgemm.cc
shz0116: Done. Also format the file using clang-format.

.github/workflows/fbgemmci.yml
shz0116: Yes.

src/Fbgemm.cc
shz0116: Done.

PR #285, opened 2020-02-04
enable github action build/test on windows

shz0116:
Summary: Enable github action on Windows for both shared/static build/test.

Differential Revision: D19706329


peterjc123:
Great work!

PR #316, opened 2020-03-14
run quantized model with 32bit lib, fix 32bit/64bit compatibility

menchunlei:
I have built 32bit lib on window10, but inference quantized model fail. while run non-quantized model successfully.
First, i fix errors when building, as follow:
(1)
error:
fbgemm\src\OptimizedKernelsAvx2.cc(126): error C3861: “_mm256_extract_epi64”: undefined
(2)
error:
src\QuantUtilsAvx2.cc(270): error C3861: “_mm256_extract_epi64”: undefined

the PR could fix the error, but run quantized model fail.

Second, i think there are some  64bit/32bit compatibility，Please give me some suggestion!!

Pytorch version: v1.3.1,  and compile 32bit lib on windows10
fbgemm version: master branch

### DIFF
--- a/src/OptimizedKernelsAvx2.cc
+++ b/src/OptimizedKernelsAvx2.cc
@@ -122,6 +122,7 @@ void transpose_8rows(
     _mm_storel_epi64(
         reinterpret_cast<__m128i*>(dst + (j + 0) * ld_dst),
         _mm256_castsi256_si128(y0));
+#if defined(_M_X64)
     *reinterpret_cast<int64_t*>(dst + (j + 1) * ld_dst) =
         _mm256_extract_epi64(y0, 1);
     _mm_storel_epi64(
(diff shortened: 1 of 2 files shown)


src/OptimizedKernelsAvx2.cc
```diff
@@ -191,6 +192,77 @@ void transpose_8rows(
         _mm256_extract_epi64(y7, 2);
     *reinterpret_cast<int64_t*>(dst + (j + 31) * ld_dst) =
         _mm256_extract_epi64(y7, 3);
+#else
+    *reinterpret_cast<int32_t*>(dst + (j + 1) * ld_dst) =
+        _mm256_extract_epi32(y0, 1);
```
jspark1105: You need to implement _mm256_extract_epi64 with multiple _mm256_extract_epi32 instructions.

*reinterpret_cast<int32_t*>(dst + (j + 1) * ld_dst) =
        _mm256_extract_epi32(y0, 2);
*reinterpret_cast<int32_t*>(dst + (j + 1) * ld_dst + 4) =
        _mm256_extract_epi32(y0, 3);

Please apply a similar to the remaining part.

BTW, we don't plan to support 32-bit architecture and IIRC PyTorch also doesn't officially support it.
So even if it works we're not going to merge the PR. Just wanted to be clear up front.

menchunlei: Hi @jspark1105 
 
Thanks for your suggestion.  
I change from:
reinterpret_cast<int32_t>(dst + (j + 1) * ld_dst + 4) =
_mm256_extract_epi32(y0, 3);
to:
reinterpret_cast<int32_t>(dst + (j + 1) * ld_dst + 1) =
_mm256_extract_epi32(y0, 3);

But forward quantized model in 32bit fail too,  i think there are some compatibilities and will continue debug.

Thanks anyway!

PR #342, opened 2020-04-08
Open sourcing fbgemm_fp16 ops

yinghai:
Summary:
So that we can let vendor use this as a reference for fp16 emulated ops.

Will modify the dependent TARGETS and CMakefiles.

Differential Revision: D20911460


yinghai:
@dskhudia How do I debug windows failure?

dskhudia:
> @dskhudia How do I debug windows failure?

There is an internal doc on this. I'll send you the link.

PR #349, opened 2020-04-10
Move `transpose_simd` to TransposeUtils.cc

malfet:
Summary:
Create `fbgemm_base` filelist, which solves circular symbol dependency issue during bazel compilation.
Bazel (or rather gold linker) does not allow circular symbols dependency between static library.
i.e. if library A have some external symbols inplemented in library B(i.e. A depends on B), then library B can not have use any symbols from library A
`Utils.cc` was one such example: `transpose_simd` used symbols from `fbgemm_avx512` and `fbgemm_avx2`, while both `fbgemm_avx[2|512]` libs depend on `fbgemmAlignedAlloc` defined in `Utils.cc`

Reviewed By: dzhulgakov

Differential Revision: D20956364


z-a-f:
Reverting as it broke `pytorch_bazel_test`

PR #370, opened 2020-05-11
Float16 rowwise sparse adagrad with stochastic rounding

uyongw:
Summary:

Add FP16 weights support to fused rowwise sparse adagrad.
- FP16 weights update use stochastic rounding by default
   * use_stochastic_rounding=true | false
- Extend existing APIs on both REF and JIT path and keep API backward compatibility;
   * REF: rowwise_sparse_adagrad_fused_ref
    * JIT: GenerateRowWiseSparseAdaGradFused
- Both API are thread-safe; a 4x64-bytes per-thread random buffer is maintained on both JIT and REF implementation; 
- Random number generating based on small-state generators xoshiro128++;
- Both AVX2 and AVX512 are implemented in JIT kernel

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,7 @@ if(MSVC)
   target_compile_options(fbgemm_avx2 PRIVATE "/arch:AVX2")
   target_compile_options(fbgemm_avx512 PRIVATE "/arch:AVX512")
 else(MSVC)
+  target_compile_options(fbgemm_generic PRIVATE "-mf16c")
   target_compile_options(fbgemm_avx2 PRIVATE
     "-m64" "-mavx2" "-mf16c" "-mfma" "-masm=intel")
   target_compile_options(fbgemm_avx512 PRIVATE
--- a/bench/RowwiseAdagradFusedBenchmark.cc
+++ b/bench/RowwiseAdagradFusedBenchmark.cc
@@ -109,9 +109,9 @@ void run_benchmark(
                    (use_fp16_weights ? sizeof(float16) : sizeof(float)) * 2 +
                    (use_32_bit_indices ? 4 : 8)) +
     batch_size * (embedding_dim * sizeof(float) + sizeof(int));
-  // FIXME: float16 is counted as float for effective byte loading
   double bytes_padded =
-    lengths_sum * (((embedding_dim * sizeof(float) + 63) / 64 + 1) * 64 * 2 +
+    lengths_sum * (((embedding_dim * (use_fp16_weights ? sizeof(float16) :
+                                      sizeof(float)) + 63) / 64 + 1) * 64 * 2 +
                    (use_32_bit_indices ? 4 : 8)) +
     batch_size * (embedding_dim * sizeof(float) + sizeof(int));
 
--- a/bench/RowwiseAdagradFusedBenchmark.cc
+++ b/bench/RowwiseAdagradFusedBenchmark.cc
@@ -45,14 +45,14 @@ void run_benchmark(
     int num_rows,
     int embedding_dim,
     int average_len,
+    bool use_fp16_weights,
     bool use_32_bit_indices = false,
     bool prefetch = false) {
   vector<char> llc(64L * 1024L * 1024L, 1.0);
   vector<float> g(batch_size * embedding_dim); // gradients
   vector<float> h(num_rows); // input momentums
   vector<float> w(num_rows * embedding_dim); // input params
-  vector<float> h_ref(h.size());
-  vector<float> w_ref(w.size());
+  vector<float16> w_fp16(w.size()); // input params
 
   default_random_engine generator;
   // normal_distribution<float> h_w_distribution;
@@ -62,10 +62,10 @@ void run_benchmark(
     g[i] = 4 + i; // h_w_distribution(generator);
   }
   for (int i = 0; i < h.size(); ++i) {
-    h_ref[i] = h[i] = 2 + i; // h_w_distribution(generator);
+    h[i] = 2 + i; // h_w_distribution(generator);
   }
   for (int i = 0; i < w.size(); ++i) {
-    w_ref[i] = w[i] = 3 + i; // h_w_distribution(generator);
+    w[i] = 3 + i; // h_w_distribution(generator);
   }
 
   // Generate lengths
@@ -104,47 +104,81 @@ void run_benchmark(
   constexpr int NUM_ITER = 10;
   // Only counts the number of bytes for reading embedding table and ignore
   // others. Should be good enough as long as embdding_dim is big enough.
-  double bytes = lengths_sum *
-          ((embedding_dim + 1) * sizeof(float) * 2 +
-           (use_32_bit_indices ? 4 : 8)) +
-      batch_size * (embedding_dim * sizeof(float) + sizeof(int));
-  double bytes_padded = lengths_sum *
-          (((embedding_dim * sizeof(float) + 63) / 64 + 1) * 64 * 2 +
-           (use_32_bit_indices ? 4 : 8)) +
-      batch_size * (embedding_dim * sizeof(float) + sizeof(int));
+  double bytes =
+    lengths_sum * ((embedding_dim + 1) *
+                   (use_fp16_weights ? sizeof(float16) : sizeof(float)) * 2 +
+                   (use_32_bit_indices ? 4 : 8)) +
+    batch_size * (embedding_dim * sizeof(float) + sizeof(int));
+  // FIXME: float16 is counted as float for effective byte loading
+  double bytes_padded =
+    lengths_sum * (((embedding_dim * sizeof(float) + 63) / 64 + 1) * 64 * 2 +
+                   (use_32_bit_indices ? 4 : 8)) +
+    batch_size * (embedding_dim * sizeof(float) + sizeof(int));
 
-  auto kernel_i32 = GenerateRowWiseSparseAdaGradFused<int32_t>(
-      embedding_dim, prefetch ? 16 : 0);
-  auto kernel_i64 = GenerateRowWiseSparseAdaGradFused<int64_t>(
-      embedding_dim, prefetch ? 16 : 0);
+  auto kernel_i32 = GenerateRowWiseSparseAdaGradFused
+    <int32_t, int32_t, float>(embedding_dim, prefetch ? 16 : 0);
+  auto kernel_i64 = GenerateRowWiseSparseAdaGradFused
+    <int64_t, int32_t, float>(embedding_dim, prefetch ? 16 : 0);
+  auto kernel_fp16_i32 = GenerateRowWiseSparseAdaGradFused
+    <int32_t, int32_t, float16>(embedding_dim, prefetch ? 16 : 0);
+  auto kernel_fp16_i64 = GenerateRowWiseSparseAdaGradFused
+    <int64_t, int32_t, float16>(embedding_dim, prefetch ? 16 : 0);
 
   for (bool flush_cache : {false, true}) {
     double t = measureWithWarmup(
         [&]() {
-          if (use_32_bit_indices) {
-            kernel_i32(
-                batch_size,
-                lengths_sum,
-                num_rows,
-                w.data(),
-                g.data(),
-                h.data(),
-                indices_32.data(),
-                lengths.data(),
-                epsilon,
-                lr);
+          if (use_fp16_weights) {
+            if (use_32_bit_indices) {
+              kernel_fp16_i32(
+                  batch_size,
+                  lengths_sum,
+                  num_rows,
+                  w_fp16.data(),
+                  g.data(),
+                  h.data(),
+                  indices_32.data(),
+                  lengths.data(),
+                  epsilon,
+                  lr);
+            } else {
+              kernel_fp16_i64(
+                  batch_size,
+                  lengths_sum,
+                  num_rows,
+                  w_fp16.data(),
+                  g.data(),
+                  h.data(),
+                  indices.data(),
+                  lengths.data(),
+                  epsilon,
+                  lr);
+            }
           } else {
-            kernel_i64(
-                batch_size,
-                lengths_sum,
-                num_rows,
-                w.data(),
-                g.data(),
-                h.data(),
-                indices.data(),
-                lengths.data(),
-                epsilon,
-                lr);
+            if (use_32_bit_indices) {
+              kernel_i32(
+                  batch_size,
+                  lengths_sum,
+                  num_rows,
+                  w.data(),
+                  g.data(),
+                  h.data(),
+                  indices_32.data(),
+                  lengths.data(),
+                  epsilon,
+                  lr);
+            } else {
+              kernel_i64(
+                  batch_size,
+                  lengths_sum,
+                  num_rows,
+                  w.data(),
+                  g.data(),
+                  h.data(),
+                  indices.data(),
+                  lengths.data(),
+                  epsilon,
+                  lr);
+            }
           }
         },
         NUM_WARMUP,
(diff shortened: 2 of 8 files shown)


src/RefImplementations.cc
```diff
@@ -1178,16 +1247,75 @@ int rowwise_sparse_adagrad_fused_ref(
...
+
+        if (isFloat16w && use_stochastic_rounding) {
+          if (sr_idx == 0) {
+            for (int v = 0; v < vlen; ++v) {
+              R[v] = rnd128_next(v, vlen);
+              r[v] = (R[v] << 24) >> 19;
+            }
+          } else if (sr_idx == 1) {
+            for (int v = 0; v < vlen; ++v)
+              r[v] = ((R[v] >> 8) << 24) >> 19;
```
jspark1105, line 1286: Maybe we can make shifting logics more consistent?
for sr_idx == 1
(R[v] << 16) >> 19
for sr_idx == 2
(R[v] << 8) >> 19
for sr_idx == 3
R[v] >> 19

src/RefImplementations.cc
```diff
@@ -1178,16 +1247,75 @@ int rowwise_sparse_adagrad_fused_ref(
...
+          if (isFloat16w) {
+            union {
+              float w_f32;
+              uint32_t w_i32;
+            };
+            w_f32 = cpu_half2float(w_[j]);
+            w_f32 = std::fma(float_step, g_[j], w_f32);
+            if (use_stochastic_rounding)
+              w_i32 += r[v];
+            w_[j] = cpu_float2half_rn(w_f32);
```
jspark1105, line 1307: Since we added positive random numbers we need to round down using _cvtps_ph(w_f32, _MM_FROUND_TO_ZERO | _MM_FROUND_NO_EXC)

jspark1105:
Thanks so much @uyongw for this valuable contribution. I added comments except to RowWiseSparseAdagradFused.cc . I'll go over RowWiseSparseAdagradFused.cc soon.

bench/RowwiseAdagradFusedBenchmark.cc
```diff
@@ -104,47 +104,81 @@ void run_benchmark(
...
-  auto kernel_i64 = GenerateRowWiseSparseAdaGradFused<int64_t>(
-      embedding_dim, prefetch ? 16 : 0);
+  double bytes =
+    lengths_sum * ((embedding_dim + 1) *
+                   (use_fp16_weights ? sizeof(float16) : sizeof(float)) * 2 +
+                   (use_32_bit_indices ? 4 : 8)) +
+    batch_size * (embedding_dim * sizeof(float) + sizeof(int));
+  // FIXME: float16 is counted as float for effective byte loading
+  double bytes_padded =
+    lengths_sum * (((embedding_dim * sizeof(float) + 63) / 64 + 1) * 64 * 2 +
```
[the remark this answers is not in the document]
uyongw: Sure I will make it the same.

src/RefImplementations.cc
uyongw: Yes it can be so but it may need an additional mask to clear other not used bits. But you are right we can still make it more consistent in REF path. Perhaps we can just use the equivalent logic below:

```
case sr_idx:
0: (R[v] && 0xFFU) << 5
1: (R[v] & 0xFF00U) >> 3
2: (R[v] & 0xFF0000U) >> 11
3: (R[v] & 0xFF000000U) >> 19
```

In JIT path the above logic requires additional vector registers (and additional mask loads in AVX2) this is why the shift is used.

src/RefImplementations.cc
uyongw: Good catch. I will update it.

src/RowWiseSparseAdagradFused.cc
```diff
@@ -63,45 +65,58 @@ class GenRowWiseSparseAdagradFused {
   // The hash depends on embedding dimension (block size), prefetch distance,
```
jspark1105, line 65: nit: can you update the comment that the first key in code cache is pointer to avx2 mask?

src/RowWiseSparseAdagradFused.cc
```diff
@@ -123,10 +138,10 @@ typename ReturnFunctionSignature<indxType, offsetType>::
         x86::Gp lengths = a->gpz(11);
         x86::Xmm epsilon(0);
         x86::Xmm lr(1);
-        x86::Gp mask_avx2 = a->gpz(12);
+        x86::Gp rand_buffer = a->gpz(12);
 
-        // reuse mask_avx2 because mask_avx2 is used only at the beginning
-        x86::Gpd lengths_R = a->gpz(12).r32();
+        // FP32 weights does not need rand_buffer
+        x86::Gpd lengths_R = areWeightsFp16 ? a->zbx().r32() : a->gpz(12).r32();
```
jspark1105, line 144: Maybe to simplify we always use r12 for lengths_R and rand_buffer can use zax? zax is a caller-saved register like zcx and zdx, so no need to set dirty flag separately for areWeightsFp16 case as in line 184.

src/RowWiseSparseAdagradFused.cc
```diff
@@ -210,18 +231,64 @@ typename ReturnFunctionSignature<indxType, offsetType>::
...
+          if (instSet == inst_set_t::avx2) {
+            a->vmovdqa(S0_vreg.ymm(), x86::dword_ptr(rand_buffer));
+            a->vmovdqa(S1_vreg.ymm(), x86::dword_ptr(rand_buffer,
+                                                 1 * vlen * sizeof(uint32_t)));
+            a->vmovdqa(S2_vreg.ymm(), x86::dword_ptr(rand_buffer,
+                                                 2 * vlen * sizeof(uint32_t)));
+            a->vmovdqa(S3_vreg.ymm(), x86::dword_ptr(rand_buffer,
+                                                 3 * vlen * sizeof(uint32_t)));
+          } else { // AVX512
+            a->vmovdqa32(S0_vreg, x86::dword_ptr(rand_buffer));
```
jspark1105, line 269: Can we guarantee that rand_buffer is cache line aligned? Also not sure if we need if / else for avx2 here. We can just use vmovups?

src/RowWiseSparseAdagradFused.cc
```diff
@@ -210,18 +231,64 @@ typename ReturnFunctionSignature<indxType, offsetType>::
...
         if (remainder) {
           if (instSet == inst_set_t::avx2) {
             src_vreg = vec_reg_t(first_available_vec_reg_id);
             ++first_available_vec_reg_id;
 
             mask_vreg = x86::Ymm(first_available_vec_reg_id);
             ++first_available_vec_reg_id;
-
+            // Use scratchReg1 as temp
+            a->mov(scratchReg1, asmjit::imm(mask_avx2));
```
jspark1105, line 287: Passing mask_avx2 address is a clever trick!

src/RowWiseSparseAdagradFused.cc
```diff
@@ -398,22 +462,19 @@ typename ReturnFunctionSignature<indxType, offsetType>::
...
-          a->imul(scratchReg2, static_cast<asmjit::Imm>(sizeof(float)));
         }
 
         a->add(indices, static_cast<asmjit::Imm>(sizeof(indxType)));
 
-        a->imul(scratchReg1, static_cast<asmjit::Imm>(sizeof(float)));
-
         if (prefetch) {
-          a->prefetchw(x86::dword_ptr(h, scratchReg2));
+          a->prefetchw(x86::dword_ptr(h, scratchReg2, 2));
```
jspark1105, line 470: +1

src/RowWiseSparseAdagradFused.cc
```diff
@@ -398,22 +462,19 @@ typename ReturnFunctionSignature<indxType, offsetType>::
...
         }
         // load h
-        a->movss(float_step_xmm, x86::dword_ptr(h, scratchReg1));
+        a->movss(float_step_xmm, x86::dword_ptr(h, scratchReg1, 2));
```
jspark1105, line 473: +1

src/RowWiseSparseAdagradFused.cc
```diff
@@ -398,22 +462,19 @@ typename ReturnFunctionSignature<indxType, offsetType>::
...
         // *h + final_sum
         a->addss(float_step_xmm, partial_sum_xmm);
         // store h
-        a->movss(x86::dword_ptr(h, scratchReg1), float_step_xmm);
+        a->movss(x86::dword_ptr(h, scratchReg1, 2), float_step_xmm);
```
jspark1105, line 477: +1

src/RowWiseSparseAdagradFused.cc
uyongw: Yes the rand_buffer is aligned as alignas(64). But I think unlinged load/store is safer. 

AVX2 does not have a store with int32 granularity (there is vpmaskmovd though that needs mask). And vmovups is for float point. But I guess the behavior is pretty much same.

uyongw: Just checked, vmovdqa is faster than vmovups on BDW/HSW/IVB:

**vmovdqa**
Architecture	Latency	Throughput (CPI)
Icelake	7	0.5
Skylake	0	0.5
Broadwell	1	0.25
Haswell	1	0.25
Ivy Bridge	1	0.5

**vmovups**
Architecture	Latency	Throughput (CPI)
Icelake	7	0.5
Skylake	0	0.5
Broadwell	1	0.5
Haswell	1	0.5
Ivy Bridge	1	1

And depending on the micro-architecture, there might be data bypass delays:
"On the Core2, there is an extra latency of one clock cycle when the output of a µop in the
integer unit is used as input for a µop in the floating point unit, or vice versa."
Page 113 of https://www.agner.org/optimize/microarchitecture.pdf

src/RefImplementations.cc
```diff
@@ -16,11 +16,65 @@
...
+  };
+
+  if (!g_rnd128_initialized) {
+    // Initialize rand buffer with uniq values per thread
+    uint64_t h0 = std::hash<std::thread::id>{}(std::this_thread::get_id());
+    for (auto i = 0; i < 4; ++i) {
+      // Use thread hash as seed
+      g_rnd128_buffer[i * VLEN_MAX] = rnd128_init_next(h0);
+      uint64_t h1 = g_rnd128_buffer[i * VLEN_MAX];
+      for (auto v = 1; v < VLEN_MAX; ++v) {
```
jspark1105, line 53: Can we use a different variable like j? Getting local variable shadowing warning.

src/RefImplementations.cc
```diff
@@ -1178,16 +1247,75 @@ int rowwise_sparse_adagrad_fused_ref(
...
+      // |       |        |   xxxxx|xxx     |
+      //  31      23       15       7      0
+      //
+      // Half float has 10 bits of mantissa, and float has 23, we are shifting
+      // the bits to cover the region where half floats can't represent data.
+      // This is bit 13-23 of the mantissa of fp32.
+      // This will be effectively adding a random variable of [0,1]
+
+      for (int n = 0; n < nvec; ++n) {
+        int len = (n == nvec - 1) ? rem : vlen;
```
jspark1105, line 1275: Can we use a different variable name than len to avoid local variable shadowing warning?

src/RefImplementations.cc
```diff
@@ -1129,20 +1183,35 @@ int rowwise_sparse_adagrad_ref(
...
+  constexpr bool isFloat16w = std::is_same<float16, DataType>::value;
+  // TODO: warning on vector-size not 8/16
+  int vlen = emu_vector_size;
+  // Local random buffer to emulate SIMD vector
+  // R: generated 32bit base random numbers
+  // r: extracted 8-bit for rounding
+  uint32_t *R = nullptr, *r = nullptr;
+  if (isFloat16w && use_stochastic_rounding) {
+    // Random vector buffer for stochastic rounding
+    R = new uint32_t[vlen];
```
jspark1105, line 1211: We do early return for error cases, so hard to make sure there's no leak. Can we use vector for R and r? I believe the CI test error is because of this.

src/RowWiseSparseAdagradFused.cc
jspark1105: This is a great point. Let me check if we have other cases we should use vmovdqa instead of vmovups.

src/RefImplementations.cc
uyongw: Sorry this this . Thanks reporting this.

src/RefImplementations.cc
uyongw: Thanks again!

src/RefImplementations.cc
uyongw: Good catch. Yes I realized this also and I am using a stack array instead.

jspark1105: Unfortunately MSVC doesn't like stack array with variable length. This is a reference code so using vector shouldn't be a problem.

uyongw: Oh I am using R[VLEN_MAX] where VLEN_MAX is a constexpr of 16... it will waste 32 bytes in case of avx2 emu... is that okay?

uyongw: This is what I am going to do:
```
-  uint32_t *R = nullptr, *r = nullptr;
-  if (isFloat16w && use_stochastic_rounding) {
-    // Random vector buffer for stochastic rounding
-    R = new uint32_t[vlen];
-    r = new uint32_t[vlen];
+  constexpr int VLEN_MAX = 16;
+  uint32_t R[VLEN_MAX], r[VLEN_MAX];
+  int vlen = emu_vector_size;
+  if (vlen != 8 && vlen != 16) {
+    // Raise error as it may cause buffer overflow
+    cout << "Not supported emu_vector_size: " << emu_vector_size << endl;
+    return false;
   }
```

jspark1105: Sorry missed length is compile time constant. Then there shouldn't be an issue.

uyongw:
An observation is the FP16/SR with AVX512 JIT code will be faster than AVX2 on SKX. We could probably use AVX512 as default for FP16/SR kernel.

src/RowWiseSparseAdagradFused.cc
dskhudia: I removed such usages in https://github.com/pytorch/FBGEMM/pull/368 for integer code. Still need to land this PR though.

src/RefImplementations.cc
```diff
@@ -18,7 +18,13 @@
 #include <numeric>
 #include <thread>
 #include <iostream>
-#include <immintrin.h>
+
+// for _cvtss_sh/_cvtsh_ss
+#ifdef _MSC_VER
+# include <intrin.h>
+#else
+# include <x86intrin.h>
```
jspark1105, line 26: Maybe we can keep using immintrin.h for ! _MSC_VER

src/RefImplementations.cc
```diff
@@ -18,7 +18,13 @@
...
+#endif
```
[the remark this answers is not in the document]
jspark1105: The way macro is used looks fine to me. We can make another iteration to clean up if needed. The PR looks good to me. Let me clean somethings up like a few cosmetic issues and will commit.

uyongw:
@jspark1105 Thank you for taking the time for all the valuable comments.

CMakeLists.txt
```diff
@@ -111,6 +111,7 @@ if(MSVC)
   target_compile_options(fbgemm_avx2 PRIVATE "/arch:AVX2")
   target_compile_options(fbgemm_avx512 PRIVATE "/arch:AVX512")
 else(MSVC)
+  target_compile_options(fbgemm_generic PRIVATE "-mf16c")
```
malfet, line 114: Wouldn't that makes fbgemm code incompatible with SandyBridge CPUs?

uyongw: @malfet That would be true. We need perhaps a pure C implementation of _cvtss_sh/_cvtsh_ss in the REF implementation (there is one already but missing the misc. rounding handling). I can create a PR next week for this purpose if needed. @jspark1105

jspark1105: This PR has been reverted due to an illegal instruction error in PyTorch tests. Let me check if it's related and if so commit with a C implementation.

uyongw: Sorry for the introducing this issue. And thanks for the fast turnaround! @jspark1105

Issue #374, opened 2020-05-15
[Question] using FBGEMM (int8 GEMM) in a multi-threaded application


ykim362:
Hi,
I have a question about the usage of FBGEMM in a multi-threaded program.
In my test environment, I have 24 cores CPU and 20 independent threads calling FBGEMM (int8 GEMM). All the threads are doing independent jobs, so there's no parallelism for one fbgemm call.  https://github.com/marian-nmt/marian-dev/blob/master/src/tensors/cpu/fbgemm/packed_gemm.cpp#L614

In a heavy load situation, I'd expect higher than 2000% cpu utilization, but the cpu utilization is always less than 250% (seems only one thread is using fbgemm). Have you had this kind of problem before?
(FYI, when I use fp16 fbgemm, I don't have this issue. The cpu utilization is higher than 2000%.)
https://github.com/marian-nmt/marian-dev/blob/master/src/tensors/cpu/fbgemm/packed_gemm.cpp#L516

Thank you in advance!

jianyuh:
Hi @ykim362,

Want to understand your use case: 
> I have 24 cores CPU and 20 independent threads calling FBGEMM (int8 GEMM). All the threads are doing independent jobs, so there's no parallelism for one fbgemm call.

Are you running 20 instances (processes) where each instance calls FBGEMM independently (no communications), or 1 instance with 20 threads calling multi-threaded FBGEMM?

Looking at your code (https://github.com/marian-nmt/marian-dev/blob/master/src/tensors/cpu/fbgemm/packed_gemm.cpp#L614), the tid and num_threads are passed as 0 and 1 in int8 GEMM, so this is the single-thread usage. For FP16 FBGEMM (https://github.com/marian-nmt/marian-dev/blob/master/src/tensors/cpu/fbgemm/packed_gemm.cpp#L516), you are passing the tid and num_threads, though.

ykim362:
Thanks, @jianyuh .

Yes, I am running 20 instances (20 threads in one process) and each instance calls FBGEMM independently without communications. 

For fp16 FBGEMM call, even if the parallelism is enabled, openmp is disabled in all my use cases. So, they are always 0 and 1.

dskhudia:
[...]

PR #380, opened 2020-05-21
Remove an unnecessary memory allocation

ykim362:
This is removing a memory allocation when it's not used.
This passes all the tests and benchmarks.

In some use cases, memory allocation causes a thread contention. This change could relieve that condition.
https://github.com/pytorch/FBGEMM/issues/374


dskhudia:
Thanks a lot for fixing it :)

PR #409, opened 2020-08-12
split groupConv kernel from instruction specific code

YazhiGao:
Summary: split groupConv file into architecture-specifc and architecture-agnostic file

Differential Revision: D23003602


YazhiGao:
duplicate of #412

PR #482, opened 2021-01-15
Use 40MB instead of 6 MB for flushing the cache

jianyuh:
Summary: A100 has 40 MB L2 cache, while V100 has 6 MB L2 cache.

Differential Revision: D25920854


jianyuh:
Thanks Paulius Micikevicius for suggesting this change!

PR #483, opened 2021-01-19
Fix the CUB include path

jianyuh:
Summary: The cub include path should be updated for the cases when the cub package is downloaded from the web site. It should not have “/include”.

Differential Revision: D25959605


jianyuh:
Thanks Sergei Gofman for reporting this issue!

PR #487, opened 2021-01-20
Use memory read kernel to flush the cache

jianyuh:
Reviewed By: jspark1105

Differential Revision: D25949808


jianyuh:
Thanks  Paulius Micikevicius for suggesting this change!

PR #497, opened 2021-01-26
hfp8 conversion kernels

amylittleyang:
Summary:
Diff
* Add hybrid FP8 conversion kernel supporting variable exponent, mantissa, and exponent bias.
* Main usages are (1) low-precision embedding training, trying to beat row-wise int8 training; (2) all2all communication quantization in the backward pass.

Accuracy
* For uniform distribution within dynamic range, quantization error is capped at 12% relative error. Raw quantization error in general is higher than rowwise int8.
* Addition: For the following snippet that emulates weight update to embedding during training, top quantization error for FP8 is much smaller than int8: P166771923, although on the average cases int8 is more accurate.
```
A = normal[-0.00316, 0.00316] + normal[-0.0000316, 0.0000316]
B = Quantize(Quantize(normal[-0.00316, 0.00316]) + normal[-0.0000316, 0.0000316])
```

Perf
* on par with rowwise int8 kernel: P166752555

Differential Revision: D26078959

### DIFF
--- a/fbgemm_gpu/include/fbgemm_gpu/quantize_ops.cuh
+++ b/fbgemm_gpu/include/fbgemm_gpu/quantize_ops.cuh
@@ -294,5 +294,104 @@ __global__ void _bfloat16_to_float_cuda_kernel(
   }
 }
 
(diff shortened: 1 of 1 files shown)


jspark1105:
@amylittleyang can you close this PR since it's abandoned in Phabricator?

PR #500, opened 2021-01-29
Enhance Split Table Batched Embedding prefetch code.

another-pjohnson:
Summary:
Instead of prefetch() returning the lxu_cache_locations, it appends them
to a list and consumes it from the forward() call. In addition, added an
assert to make sure the lxu_cache_locations list does not grow unbounded.

Differential Revision: D26116359


jspark1105:
@another-pjohnson saw this has conflicts. Was it correctly resolved? If so, can you close this PR?

another-pjohnson:
I verified that these changes were properly synced, and are already in pytorch/FBGEMM.

PR #501, opened 2021-01-30
Reduce the number of launched threads in the backward pass

jianyuh:
Summary:
We only need to spawn `sorted_linear_indices_cumulative_run_lengths.numel()` for gridDim.x instead of `linear_indices.numel() `.

Note that the second is always bigger than the first due to run-length encoding: https://nvlabs.github.io/cub/structcub_1_1_device_run_length_encode.html).

Differential Revision: D26159054


jianyuh:
Thanks Sergei Gofman for reporting this issue!

PR #526, opened 2021-02-28
enable split_table_batched_embeddings_test runnable on CPU

jspark1105:
Summary: Fix use_cpu to True if cuda is not available

Differential Revision: D26717294


jspark1105:
Abandoned

PR #661, opened 2021-08-04
add sort based csr2csc

ashokei:


fbgemm_gpu/include/fbgemm_gpu/cpu_utils.h
```diff
@@ -0,0 +1,117 @@
...
+    Radix_Sort_Pair<T>* output = tmp_buf;
+
+    for (unsigned int pass = 0; pass < num_passes; pass++) {
+      // Step 1: compute histogram
+      for (int i = 0; i < RDX_HIST_SIZE; i++)
+        local_histogram[i] = 0;
+
+#pragma omp for schedule(static)
+      for (int64_t i = 0; i < elements_count_4; i += 4) {
+        T val_1 = input[i].first;
```
jspark1105, line 49: @ashokei have you tested if we have key and value as separate arrays? This may help vectorization. Although not entirely sure if vectorization will help this kind loop.

PR #753, opened 2021-11-10
fix bug in CPU exact rowwise adagrad for shared table

jspark1105:
Summary: A bug was introduced in https://github.com/pytorch/FBGEMM/pull/661 where the output of csr2csc has changed for unweighted case but the code path using fbgemm CPU JIT'ed kernel was not properly updated.

Differential Revision: D32304217


jspark1105:
CC @ashokei

PR #950, opened 2022-02-25
Split CPU and GPU ops in permute_pooled_embs_auto_grad

geyyer:
Summary: Separate permute_pooled_embs_auto_grad_cpu, refactor TARGETS and CMakeLists

Differential Revision: D34292611

PR #987, opened 2022-03-17
split cpu parts in permute_pooled_embedding_ops for cpu_only

RabbitWhite1:
As specified in [CMakeLists.txt](https://github.com/pytorch/FBGEMM/blob/c3a26e17e7c514041ae3d08f39d0e19063614869/fbgemm_gpu/CMakeLists.txt#L239), "src/permute_pooled_embedding_ops_gpu.cpp" will only compile when "NOT FBGEMM_CPU_ONLY", which means the method "permute_pooled_embs_auto_grad" won't be generated when --cpu_only. However, this method is used by torchrec's column_wise sharding.

The pr mentioned in https://github.com/pytorch/FBGEMM/pull/950 cannot work because of not using m.def to define permute_pooled_embs_auto_grad.


jianyuh:
cc @geyyer : please take a look at this PR.

RabbitWhite1:
Fixed duplicated m.def in [permute_pooled_embedding_ops_gpu.cpp](https://github.com/pytorch/FBGEMM/pull/987/commits/70d0bb5fda5c105af95cb44ef457f11e0f31c3b3#diff-4bf487dea6860e5b3f2e7bc594295fbb93bd2f6e07bbdf29ce0c993b3468dcdaR44)

geyyer:
@RabbitWhite1, thanks for contributing and noticing the possible issue with cpu-only build! To avoid compatibility issue we have to create an empty file permute_pooled_embs_function.h and then move the class template PermutePooledEmbsFunction there. The PR looks good, could you move permute_pooled_embedding_ops_utils.h contents to permute_pooled_embs_function.h and fix dependencies?

RabbitWhite1:
> @RabbitWhite1, thanks for contributing and noticing the possible issue with cpu-only build! To avoid compatibility issue we have to create an empty file permute_pooled_embs_function.h and then move the class template PermutePooledEmbsFunction there. The PR looks good, could you move permute_pooled_embedding_ops_utils.h contents to permute_pooled_embs_function.h and fix dependencies?

Great! I've moved and committed.

PR #1044, opened 2022-04-13
Fix python module import error of rowwise_adagrad_with_weight_decay

KanchiShimono:
Fix python `ModuleNotFoundError` when import new modules introduced by https://github.com/pytorch/FBGEMM/commit/1ececf6b249aed6b8d8af9f550df0120666cf67f.

Is it better to add step checking generated code to CI workflows?

```sh
python -c "import fbgemm_gpu.split_embedding_codegen_lookup_invokers"
```


KanchiShimono:
@jianyuh 
Thanks for reply and following up the test coverage!
This PR will be resolved by #1045 and #1051, so I close this.

PR #1083, opened 2022-04-28
Add permute_pooled_embedding_modules_split to unblock TorchRec

geyyer:
Differential Revision: D36009667


geyyer:
Already fixed a while ago.

PR #1096, opened 2022-05-04
Upgrade asmjit (combined)

jianyuh:
Summary: From D35292923

Differential Revision: D36121284


powderluv:
Anychance we can revive / land this one ? Clang-14 and above require latest patches on asmjit to compile without `[-Werror,-Wbitwise-instead-of-logical] ` issues

PR #1102, opened 2022-05-05
ROCm changes

liligwu:
Enabling FBGEMM on AMD devices.


fbgemm_gpu/include/fbgemm_gpu/hipcub_namespace_postfix.cuh
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
jianyuh, line 2: Can we merge this file with `cub_namespace_postfix.cuh`?

fbgemm_gpu/include/fbgemm_gpu/hipcub_namespace_postfix.hpp
```diff
@@ -0,0 +1,21 @@
+/*
```
jianyuh, line 1: Do we need this .hpp file ? (dup with .cuh file above)

jianyuh (commented):
Thanks for the contribution to FBGEMM_GPU! Leave some initial comments for review.

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "third_party/googletest"]
 	path = third_party/googletest
 	url = https://github.com/google/googletest
+[submodule "third_party/hipify_torch"]
```
shintaro-iwasaki, line 10: Do we need this? Has PyTorch already adopted this submodule?

fbgemm_gpu/CMakeLists.txt
```diff
@@ -209,15 +285,15 @@ set(cpp_fbgemm_files_avx2 "../src/EmbeddingSpMDMAvx2.cc"
...
 set(cpp_fbgemm_files_avx512 "../src/EmbeddingSpMDMAvx512.cc")
-
-set_source_files_properties(
-  ${cpp_fbgemm_files_avx512}
-  PROPERTIES COMPILE_OPTIONS
-             "-mavx2;-mf16c;-mfma;-mavx512f;-mavx512bw;-mavx512dq;-mavx512vl")
-
-set(cpp_fbgemm_files ${cpp_fbgemm_files_normal} ${cpp_fbgemm_files_avx2}
-                     ${cpp_fbgemm_files_avx512})
+if(USE_CUDA)
```
shintaro-iwasaki, line 290: Is AVX512 related to CUDA?

fbgemm_gpu/CMakeLists.txt
```diff
@@ -330,7 +435,9 @@ if(NVML_LIB_PATH)
   target_link_libraries(fbgemm_gpu_py ${NVML_LIB_PATH})
 endif()
 target_include_directories(fbgemm_gpu_py PRIVATE ${TORCH_INCLUDE_DIRS})
-set_property(TARGET fbgemm_gpu_py PROPERTY CXX_STANDARD 17)
+if(USE_CUDA)
+  set_property(TARGET fbgemm_gpu_py PROPERTY CXX_STANDARD 17)
+endif()
```
shintaro-iwasaki, line 440: Is this a CUDA-specific property?

fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
```diff
@@ -797,8 +797,10 @@ def uvm(
             offsets = torch.tensor(([0] + np.cumsum(lengths).tolist())).int().cuda()
             per_sample_weights = None
             if weighted:
-                assert (this_rs_uvm_weights := rs_uvm[2]) is not None
-                assert (this_rs_gpu_weights := rs_gpu[2]) is not None
+                this_rs_uvm_weights = rs_uvm[2]
+                assert this_rs_uvm_weights is not None
+                this_rs_gpu_weights = rs_gpu[2]
+                assert this_rs_gpu_weights is not None
```
shintaro-iwasaki, line 803: Could you explain why this change is needed for ROCm?

fbgemm_gpu/codegen/embedding_backward_code_generator.py
```diff
@@ -1316,3 +1316,4 @@ def main() -> None:
 
 if __name__ == "__main__":
     main()
+    # hipify_gen()
```
shintaro-iwasaki, line 1319: Please do not leave a meaningless comment.

fbgemm_gpu/codegen/embedding_backward_split_template.cu
```diff
@@ -963,7 +968,6 @@ split_embedding{{ "_nobag" if nobag else "" }}_backward_codegen_{{ optimizer }}_
             // over 48 KB per block are architecture-specific, as such they
             // must use dynamic shared memory (rather than statically sized
             // arrays) and require an explicit opt-in using cudaFuncSetAttribute()".
-
```
shintaro-iwasaki, line 966: Would you mind removing such unnecessary changes that are left in the code? It makes us easy to review the code (also keep the diff minimized).

fbgemm_gpu/include/fbgemm_gpu/embedding_backward_template_helpers.cuh
```diff
@@ -10,9 +10,17 @@
 #include <ATen/TensorUtils.h>
 #include <ATen/core/TensorAccessor.h>
 #include <ATen/cuda/CUDAContext.h>
+#if !defined(NEW_GENERATOR_PATH)
+#include <ATen/CUDAGeneratorImpl.h>
```
jianyuh, line 14: Can we remove `NEW_GENERATOR_PATH` and only use `ATen/cuda/CUDAGeneratorImpl.h` for PyTorch 1.11+ ?

fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
liligwu: > Could you explain why this change is needed for ROCm?

ROCm is on Python 3.7.

fbgemm_gpu/src/cumem_utils.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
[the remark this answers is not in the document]
liligwu: It's corrected.

fbgemm_gpu/run_all.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
liligwu: Removing this script in the next commit since it is not in the upstream.

fbgemm_gpu/include/fbgemm_gpu/hipcub_namespace_postfix.cuh
liligwu: These `hipcub_namespace_postfix.cuh`, `hipcub_namespace_postfix.hpp` are accidentally added in git. They should be removed. Thanks for reminding me.

fbgemm_gpu/include/fbgemm_gpu/hipcub_namespace_postfix.hpp
liligwu: The same as above.

.gitmodules
liligwu: Yes, we are moving to hipify-torch repo, even for PyTorch this change is under review.

fbgemm_gpu/CMakeLists.txt
liligwu: AMD CPU has no support for avx512, and the compiler will complain if adding the compilation flag.

fbgemm_gpu/codegen/embedding_backward_code_generator.py
liligwu: It's removed.

fbgemm_gpu/codegen/embedding_backward_split_template.cu
liligwu: The empty line is taken back.

fbgemm_gpu/include/fbgemm_gpu/embedding_backward_template_helpers.cuh
liligwu: We moved to ROCm5.1 and PyTorch1.11. These branches were deleted.

fbgemm_gpu/CMakeLists.txt
liligwu: ROCm c++ standard is set on https://github.com/ROCmSoftwarePlatform/FBGEMM/blob/eabd0a8723efff806eb3c4e8947ee754f56dfdba/fbgemm_gpu/cmake/Hip.cmake#L136

fbgemm_gpu/CMakeLists.txt
shintaro-iwasaki: In my understanding, HIP is for AMD GPUs so it is irrelevant. For example, I believe we can use AMD GPUs with an Intel Skylake CPU, which has AVX512.

At least, it is not related to CUDA. Can I ask you to fix properly (e.g., detecting the availability of AVX-512 instructions)?

(@jianyuh Please teach me how FBGEMM should handle AVX-512.)

fbgemm_gpu/CMakeLists.txt
shintaro-iwasaki: Can't ROCm also use C++17 (as CUDA does)? Or is there any reason to stick to C++14? I am not sure if FBGEMM stay compilable with C++14 in the future.

liligwu: > Can't ROCm also use C++17 (as CUDA does)? Or is there any reason to stick to C++14? I am not sure if FBGEMM stay compilable with C++14 in the future.

We are using c++ 17, please see the link above.

fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
shintaro-iwasaki: Maybe I just don't understand Python, but
- Does ROCm depend on a specific Python version? In my understanding, it is irrelevant.
- Does the original grammar depend on a specific Python version (e.g., only Python2, or Python 3.8+, or Python <3.5).
```py
assert (this_rs_uvm_weights := rs_uvm[2]) is not None
```

liligwu: ":=" is a Python 3.8 feature, see https://stackoverflow.com/a/26000366
The PyTorch upstream CI jobs for ROCm are executed with python 3.7, so all our release dockers are on 3.7.

fbgemm_gpu/CMakeLists.txt
shintaro-iwasaki: If ROCm is also using C++17, do you need to add this `if (USE_CUDA)` branch?

fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
shintaro-iwasaki: Yes maybe we should change this if FBGEMM_GPU should work with Python < 3.8,, but I am not sure about the minimum Python version for FBGEMM_GPU, but @jianyuh do you know the version requirement?

fbgemm_gpu/CMakeLists.txt
```diff
@@ -147,14 +171,30 @@ set(codegen_dependencies
     ${CMAKE_CURRENT_SOURCE_DIR}/include/fbgemm_gpu/sparse_ops_utils.h
 )
 
-add_custom_command(
-  OUTPUT ${gen_cpu_source_files} ${gen_gpu_source_files}
-         ${gen_gpu_host_source_files} ${gen_python_files}
-  COMMAND
+if(USE_ROCM)
+	execute_process(
```
shintaro-iwasaki, line 175: Can you fix the white space (I think here you are using a tab "\t").

fbgemm_gpu/CMakeLists.txt
```diff
@@ -147,14 +171,30 @@ set(codegen_dependencies
...
+		DEPENDS "${codegen_dependencies}")
+
+  set(header_include_dir
+      ${CMAKE_CURRENT_SOURCE_DIR}/include
+      ${CMAKE_CURRENT_SOURCE_DIR}/src
+      ${CMAKE_CURRENT_SOURCE_DIR})
+
+  hipify(CUDA_SOURCE_DIR ${PROJECT_SOURCE_DIR} HEADER_INCLUDE_DIR ${header_include_dir})
+else()
+	add_custom_command(
```
shintaro-iwasaki, line 189: Ditto. Please follow the original whitespace formatting (in this case, tab -> whitespace)

fbgemm_gpu/cmake/Hip.cmake
```diff
@@ -0,0 +1,176 @@
...
+ELSE()
+  SET(HIPSPARSE_PATH $ENV{HIPSPARSE_PATH})
+ENDIF()
+
+# THRUST_PATH
+IF(DEFINED ENV{THRUST_PATH})
+  SET(THRUST_PATH $ENV{THRUST_PATH})
+ELSE()
+  SET(THRUST_PATH ${ROCM_PATH}/include)
+ENDIF()
```
shintaro-iwasaki, line 67: nit: why only this is the opposite? Can you make a condition as follows to unify the condition?
```cmake
IF(NOT DEFINED ENV{THRUST_PATH})
```

fbgemm_gpu/codegen/embedding_forward_quantized_split_template.cu
```diff
@@ -505,14 +509,13 @@ __global__ __launch_bounds__(kMaxThreads) void int_nbit_split_embedding_codegen_
...
-            assert(false);
-            if (__any(found)) {
+            if (__any_sync(subwarp_mask, found)) {
 #else
             if (__any_sync(subwarp_mask, found)) {
 #endif
                 break;
 #ifdef __HIP_PLATFORM_HCC__
-            } else if (__any(empty)) {
+            } else if (__any_sync(subwarp_mask, empty)) {
```
shintaro-iwasaki, line 518: Now both HIP and CUDA versions look the same (`__any_sync(subwarp_mask, empty`).

Can we remove this `#ifdef`?

fbgemm_gpu/include/fbgemm_gpu/enum_utils.h
```diff
@@ -15,9 +15,11 @@ namespace fbgemm_gpu {
 
 #define FBGEMM_GPU_ENUM_CREATE_TAG(module_name)                                \
   struct fbgemm_gpu_enum_tag_##module_name {};                                 \
-  extern template enum_registration<struct fbgemm_gpu_enum_tag_##module_name>* \
+  template <> enum_registration<struct fbgemm_gpu_enum_tag_##module_name>*     \
       enum_registration<                                                       \
-          struct fbgemm_gpu_enum_tag_##module_name>::registration_list;
+          struct fbgemm_gpu_enum_tag_##module_name>::registration_list;        \
+  extern template class enum_registration<                                     \
+      struct fbgemm_gpu_enum_tag_##module_name>;
```
shintaro-iwasaki, line 22: Would you mind explaining why this change is needed for ROCm?

fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh
```diff
@@ -625,7 +625,11 @@ template <typename T, int ReduceWidth = kWarpSize>
 DEVICE_INLINE T warpReduceAllSum(T val) {
 #pragma unroll
   for (int mask = ReduceWidth / 2; mask > 0; mask >>= 1) {
+#ifdef __HIP_PLATFORM_HCC__
+    val += __shfl_xor(val, mask);
+#else
     val += shfl_xor(val, mask);
+#endif
```
shintaro-iwasaki, line 632: `shfl_xor()` is a wrapper to support both HIP and CUDA.
https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh#L614-L621

Do you still need this branch? If you need it, is it possible to update that wrapper?

fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh
```diff
@@ -1212,6 +1216,9 @@ DEVICE_INLINE float_16 make_zero_float_16() {
 
 __forceinline__ __device__ __half2
 hfma2(const __half2 a, const __half2 b, const __half2 c) {
+#ifdef __HIP_PLATFORM_HCC__
+  return __hfma2(a, b, c);
+#else
```
shintaro-iwasaki, line 1221: Can I ask you to merge the condition? For example,
```
#if (__CUDA_ARCH__ >= 530 && __CUDA_ARCH__ != 610) || defined(__HIP_PLATFORM_HCC__)
...
```

It'd be nice if you can apply this change to other changes below.

fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh
```diff
@@ -2211,4 +2223,11 @@ DEVICE_INLINE float float16_min(float_16 val) {
 #undef min
 #undef max
 
+#ifdef __HIP_PLATFORM_HCC__
+__device__ int __any_sync(uint64_t mask, int predicate) {
+  uint64_t predicate_bit_pattern = __ballot(predicate);
+  return (predicate_bit_pattern & mask) > 0;  
+}
+#endif
```
shintaro-iwasaki, line 2231: Is this what CUDA natively supports (e.g., in some CUDA headers) but HIP currently does not support? If so, I'd appreciate it if you could add a comment.

fbgemm_gpu/src/cumem_utils.cu
```diff
@@ -295,6 +324,7 @@ void uvm_cuda_mem_advise(Tensor t, int64_t cudaMemoryAdvise) {
 #endif
```
shintaro-iwasaki, line 324: First of all, do you need to create a separate function for this (i.e., can't the hipification script convert the CUDA version)?

If we don't need to, we have to maintain two functions, which would be great.
If we need to create a separate function for AMD GPUs, can you remove the outdated `__HIP_PLATFORM_HCC__` from the CUDA code above?

fbgemm_gpu/src/cumem_utils.cu
```diff
@@ -357,10 +387,18 @@ Tensor uvm_to_cpu_clone(Tensor t) {
   return cpu_clone;
 }
 
-#ifndef __HIP_PLATFORM_HCC__
-// FIXME: some advanced "cudaMemAdvise" flags are not supported by HIP.
 FBGEMM_GPU_ENUM_GLOGAL(uvm)
-
+#ifdef __HIP_PLATFORM_HCC__
+// FIXME: some advanced "cudaMemAdvise" flags are not supported by HIP.
+FBGEMM_GPU_ENUM_REGISTER_START(uvm, hipMemoryAdvise){
```
shintaro-iwasaki, line 393: 1. Is this `FIXME` still needed? It looks that all the flags are supported by HIP (at least listed).
2. Can't we rely on hipification script? It looks that the only difference is `hipMemoryAdvise` vs `cudaMemoryAdvise`.

fbgemm_gpu/src/quantize_ops.cu
```diff
@@ -109,7 +109,9 @@ __global__ inline void _float_to_fused8bitrowwise_cuda_kernel(
 template <typename T>
 __device__ inline __attribute__((always_inline)) T
 quantize_ops_shfl_xor(const T val, int laneMask, int width) {
-#if CUDA_VERSION >= 9000
+#ifdef __HIP_PLATFORM_HCC__
```
shintaro-iwasaki, line 112: Can we merge this condition to avoid duplicated `__shfl_xor(val, laneMask, width);`?

fbgemm_gpu/src/split_embeddings_cache_cuda.cu
```diff
@@ -438,8 +438,7 @@ __global__ __launch_bounds__(kMaxThreads) void lru_cache_find_uncached_kernel(
   // FIXME: __any_sync with mask isn't supported by HIP yet.
   // See https://fburl.com/fvy7j0lq for the similar context.
   // assert false here with https://fburl.com/pfm7enw2
```
shintaro-iwasaki, line 440: It looks that `!__any_sync(0xFFFFFFFFFFFFFFFF, found)` solved the issue, but do we need to keep this `FIXME`?

fbgemm_gpu/src/split_embeddings_cache_cuda.cu
```diff
@@ -1167,8 +1166,7 @@ __global__ __launch_bounds__(kMaxThreads) void lfu_cache_find_uncached_kernel(
...
```
shintaro-iwasaki, line 1168: Ditto. Do we need to keep this `FIXME`?

fbgemm_gpu/test/test_utils.py
```diff
@@ -182,3 +186,14 @@ def cpu_and_maybe_gpu() -> st.SearchStrategy[List[torch.device]]:
...
+def skipIfRocm(reason="test doesn't currently work on the ROCm stack"):
+    def skipIfRocmDecorator(fn):
+        @wraps(fn)
+        def wrapper(*args, **kwargs):
+            if TEST_WITH_ROCM:
+                raise unittest.SkipTest(reason)
+            else:
+                fn(*args, **kwargs)
+        return wrapper
+    return skipIfRocmDecorator
```
shintaro-iwasaki, line 199: Please add a newline here.

fbgemm_gpu/src/sparse_ops.cu
```diff
@@ -2128,6 +2132,52 @@ Tensor permute102_baddbmm_permute102_cuda(
...
+  auto lda = k * batch_size;
+  auto strideA = k;
+
+  auto Ctype = HIPBLAS_R_16F;
+  auto ldc = n * batch_size;
+  auto strideC = n;
+
+  auto computeType = HIPBLAS_R_32F;
+
+  auto result = hipblasGemmStridedBatchedEx(
```
shintaro-iwasaki, line 2153: Just out of curiosity: why only this function is supported by hipBLAS but not rocBLAS?

fbgemm_gpu/cmake/Hip.cmake
```diff
@@ -0,0 +1,176 @@
...
+  set(ROCclr_DIR ${ROCM_PATH}/rocclr/lib/cmake/rocclr)
+  set(ROCRAND_INCLUDE ${ROCRAND_PATH}/include)
+  set(ROCM_SMI_INCLUDE ${ROCM_PATH}/rocm_smi/include)
+
+  set(FBGEMM_HIP_INCLUDE ${ROCM_PATH}/include ${FBGEMM_HIP_INCLUDE})
+  set(FBGEMM_HIP_INCLUDE ${hip_INCLUDE_DIRS} $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}> $<INSTALL_INTERFACE:include> ${FBGEMM_HIP_INCLUDE})
+
+  hip_include_directories(${FBGEMM_HIP_INCLUDE} ${ROCRAND_INCLUDE} ${ROCM_SMI_INCLUDE})
+
+  list (APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
```
shintaro-iwasaki, line 171: I thought basically `${ROCM_PATH}` is expected, but does this line need to rely on a fixed `/opt/rocm` path?

shintaro-iwasaki (commented):
Thank you very much for your clean up @liligwu!
I added some comments.

fbgemm_gpu/cmake/Hip.cmake
liligwu: It's kept aligned.

fbgemm_gpu/codegen/embedding_forward_quantized_split_template.cu
liligwu: It's removed.

fbgemm_gpu/include/fbgemm_gpu/enum_utils.h
liligwu: When using hipcc, the compiler complains `error: explicit specialization of 'registration_list' after instantiation`. 
Based on https://stackoverflow.com/questions/54483114/specialization-of-a-template-class-static-member-variable-when-also-using-exter, one had to explicitly declare the static class member for the specialized template prior to instantiating the specialized template later.

fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh
liligwu: The branch is removed.

fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh
liligwu: Thank you for your suggestions. The conditions are combined.

fbgemm_gpu/src/cumem_utils.cu
liligwu: Thank you for your suggestions, I'm removing the branch as well as the `FIXME` comment.

fbgemm_gpu/src/quantize_ops.cu
liligwu: The conditions were combined and aligned with the ones in `fbgemm_cuda_utils.cuh`.

fbgemm_gpu/src/split_embeddings_cache_cuda.cu
liligwu: The comments are removed.

fbgemm_gpu/src/split_embeddings_cache_cuda.cu
liligwu: The comments are removed.

fbgemm_gpu/test/test_utils.py
liligwu: It's added.

fbgemm_gpu/cmake/Hip.cmake
liligwu: Good point. It's changed.

fbgemm_gpu/CMakeLists.txt
liligwu: 'hip_add_library' has a different way to set the compilation flags with `HIPCC_OPTIONS` (https://github.com/ROCmSoftwarePlatform/FBGEMM/blob/99a70e143e7ee21565d17cee2c66930ba743253e/fbgemm_gpu/CMakeLists.txt#L419). We can remove this `if (USE_CUDA)` condition but leave it as an ineffective statement. Would you like me to remove it? Thank you @shintaro-iwasaki

liligwu:
Hi @shintaro-iwasaki , I've addressed your new comments. Would you please reply to https://github.com/pytorch/FBGEMM/pull/1102#discussion_r870609756 . I appreciate your help in improving the code quality.

fbgemm_gpu/CMakeLists.txt
shintaro-iwasaki: We should minimize the gap if it does not cause a problem (e.g., error or warning). Can we remove this condition if ROCm will also use C++17?

fbgemm_gpu/src/cumem_utils.cu
shintaro-iwasaki: Just a reminder: do you need a separate function for `uvm_cuda_mem_advise()`?

fbgemm_gpu/test/test_utils.py
```diff
@@ -182,3 +186,15 @@ def cpu_and_maybe_gpu() -> st.SearchStrategy[List[torch.device]]:
...
+
```
shintaro-iwasaki, line 200: nit: I believe the problem is whitespace before the newline. Can you remove it?

src/EmbeddingSpMDM.cc
```diff
@@ -1377,6 +1377,7 @@ void compressed_indices_remap(
 
   const inst_set_t isa = fbgemmInstructionSet();
   if (isZmm(isa)) {
+#ifndef __HIP_PLATFORM_HCC__
```
shintaro-iwasaki, line 1380: I believe this change is also related to AVX-512. I might overlook the conclusion of the AVX-512 discussion, but I feel this part is irrelevant to "HIP" or "AMD GPUs".

shintaro-iwasaki:
@liligwu Thank you for your updates! Since this PR is big, I needed to (and still need to) leave nitpicking comments. Please take a look at them.

fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
jianyuh: Thanks! If the current change works for 3.7+, it looks good to me. Ideally we should use the simple syntax and avoid using the specific Python 3.8+ features.

fbgemm_gpu/src/cumem_utils.cu
liligwu: Sorry, I overlooked this comment before. 
Currently, we need to keep this `#ifdef __HIP_PLATFORM_HCC__` separated block since `hipify_torch` has some problem of converting `cudaMemoryAdvise`. I'm reaching out to the owner of `hipify_torch`.

fbgemm_gpu/CMakeLists.txt
liligwu: OK. will do.

fbgemm_gpu/test/test_utils.py
liligwu: Yeah. It's removed.

src/EmbeddingSpMDM.cc
liligwu: We cannot enable the piece of code block unless the other avx512-related problems are resolved. Otherwise, an "undefined symbol" error is thrown when python imports 

> fbgemm_gpu.

liligwu:
@shintaro-iwasaki, I addressed your yesterday's comments. Please feel free to continue suggesting the code. Thank you.

fbgemm_gpu/src/cumem_utils.cu
pruthvistony: @liligwu , 
Let get this resolved in `hipify_torch` repo.

fbgemm_gpu/CMakeLists.txt
```diff
@@ -9,13 +9,27 @@ message("${message_line}")
...
+if(EXISTS "/usr/bin/nvidia-smi")
+  message("NVIDIA GPU detected.")
+  option(USE_CUDA "Use CUDA" ON)
+  option(USE_ROCM "Use ROCm" OFF)
+elseif(EXISTS "/opt/rocm/bin/rocm-smi")
+  message("AMD GPU detected.")
+  option(USE_CUDA "Use CUDA" OFF)
+  option(USE_ROCM "Use ROCm" ON)
+else()
+  message("Unable to detect GPU vendor")
```
jianyuh, line 22: Please check the test failure on the PR: `build_gpu` and `build_cpu_only` tests. We might install fbgemm on pure CPU machines (with --cpu-only option), or we might install fbgemm GPU on a pure CPU machine to test the build correctness (installing CUDA library on the pure CPU machine but not runnable).

fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops.py
```diff
@@ -20,7 +20,18 @@
 from fbgemm_gpu.split_embedding_configs import SparseType
 from torch import Tensor, nn
 
-ASSOC = 32
+ASSOC = 32 if torch.version.hip is None else 64
+try:
```
jianyuh, line 24: Remove this to sync with the trunk?

fbgemm_gpu/test/quantize_ops_test.py
```diff
@@ -871,5 +872,39 @@ def test_quantize_and_dequantize_op_cuda_large_nrows_bf16(
             torch.testing.assert_allclose(dequantized_data_gpu.cpu(), dequantized_data)
 
 
+class TestDenseMLPQuantizationConversion(unittest.TestCase):
```
jianyuh, line 875: Remove this to sync with the trunk?

fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops.py
liligwu: The `try` block is removed.

fbgemm_gpu/CMakeLists.txt
liligwu: I've changed the logic of detecting the GPU vendor, and waiting for the test.

fbgemm_gpu/test/quantize_ops_test.py
liligwu: When enabling the test, an error is thrown `RuntimeError: No such operator fb::FloatToMSFPQuantized`. Is there any specific version of Pytorch we need to test?

fbgemm_gpu/src/cumem_utils.cu
liligwu: We've updated the mapping of `cudaMemoryAdvise` in `hipify_torch`. The `#ifdef __HIP_PLATFORM_HCC__` is removed.

fbgemm_gpu/CMakeLists.txt
liligwu: The CI passed.

fbgemm_gpu/test/quantize_ops_test.py
jianyuh: We need to use torch.ops.fbgemm.FloatToMSFPQuantized

fbgemm_gpu/include/fbgemm_gpu/enum_utils.h
jianyuh: The change in this file here caused an internal issue with the error message like `stderr: ld.lld: error: undefined symbol: fbgemm_gpu::enum_registration<fbgemm_gpu::fbgemm_gpu_enum_tag_uvm>::enum_query[abi:cxx11]()`, so it wasn't merged into the trunk. Feel free to add another PR to incorporate the changes and we can do more testings.

liligwu: > The change in this file here caused an internal issue with the error message like `stderr: ld.lld: error: undefined symbol: fbgemm_gpu::enum_registration<fbgemm_gpu::fbgemm_gpu_enum_tag_uvm>::enum_query[abi:cxx11]()`, so it wasn't merged into the trunk. Feel free to add another PR to incorporate the changes and we can do more testings.

Hi @jianyuh. Is this error occurs when importing fbgemm_gpu or when compiling the project? We don't have this issue on our side. Would you please provide more info? such that we could reproduce the error. Thank you.

shintaro-iwasaki: @liligwu Thanks. We looked at this issue. It seems that the difference comes from NVCC vs AMD Clang. You can reproduce it by running `uvm_test.py` with CUDA enabled.

https://godbolt.org/z/johx4fEME is a very simplified current code in FBGEMM (the core code is borrowed from https://stackoverflow.com/questions/54483114/specialization-of-a-template-class-static-member-variable-when-also-using-exter). This code is acceptable for GCC (and supposedly GCC-based NVCC) while not acceptable for Clang (and supposedly Clang-based AMD compiler).

I am not sure which compiler's behavior is correct in terms of C/C++, but I'd appreciate it if you could find a solution that works for both platforms.

liligwu: > @liligwu Thanks. We looked at this issue. It seems that the difference comes from NVCC vs AMD Clang. You can reproduce it by running `uvm_test.py` with CUDA enabled.
> 
> https://godbolt.org/z/johx4fEME is a very simplified current code in FBGEMM (the core code is borrowed from https://stackoverflow.com/questions/54483114/specialization-of-a-template-class-static-member-variable-when-also-using-exter). This code is acceptable for GCC (and supposedly GCC-based NVCC) while not acceptable for Clang (and supposedly Clang-based AMD compiler).
> 
> I am not sure which compiler's behavior is correct in terms of C/C++, but I'd appreciate it if you could find a solution that works for both platforms.

We'll resolve this issue in the next PR.

PR #1142, opened 2022-05-26
ROCm changes on uvm.

liligwu:
- Fixing the package import issue in uvm_test.py (4 tests skipped). 
- Changing the detection of ROCm in CMakeList.txt to `/opt/rocm`.


fbgemm_gpu/bench/quantize_ops_benchmark.py
```diff
@@ -44,6 +45,9 @@ def cli() -> None:
     num_rows=st.sampled_from([2**n for n in range(4, 10)]),
 )
 def bench(
+=======
+def bench_impl(
+>>>>>>> 33c5e06... update
```
shintaro-iwasaki, line 50: I don't think this PR is ready to review. Can you please check the whole change again?

fbgemm_gpu/include/fbgemm_gpu/enum_utils.h
```diff
@@ -15,14 +15,17 @@ namespace fbgemm_gpu {
 
 #define FBGEMM_GPU_ENUM_CREATE_TAG(module_name)                                \
   struct fbgemm_gpu_enum_tag_##module_name {};                                 \
-  extern template enum_registration<struct fbgemm_gpu_enum_tag_##module_name>* \
+  template <> enum_registration<struct fbgemm_gpu_enum_tag_##module_name>*     \
       enum_registration<                                                       \
-          struct fbgemm_gpu_enum_tag_##module_name>::registration_list;
+          struct fbgemm_gpu_enum_tag_##module_name>::registration_list;        \
+  extern template class enum_registration<                                     \
+      struct fbgemm_gpu_enum_tag_##module_name>;
```
shintaro-iwasaki, line 22: Does this modification work for CUDA side too? Can you compile and run `uvm_test.py` using CUDA?

fbgemm_gpu/test/uvm_test.py
```diff
@@ -19,13 +19,16 @@
...
     from fbgemm_gpu.test.test_utils import gpu_available, gpu_unavailable
 
 if gpu_available:
-    # pyre-ignore[21]
-    from fbgemm_gpu.uvm import cudaMemAdvise, cudaMemoryAdvise, cudaMemPrefetchAsync
+    if torch.version.hip:
+        # pyre-ignore[21]
+        from fbgemm_gpu.uvm import cudaMemAdvise, hipMemoryAdvise, cudaMemPrefetchAsync
+    else:
+        from fbgemm_gpu.uvm import cudaMemAdvise, cudaMemoryAdvise, cudaMemPrefetchAsync
```
shintaro-iwasaki, line 31: Why do we need to change only `cudaMemoryAdvise` to `hipMemoryAdvise`?
Can't we use the same symbols?

I understand that the hipification script changes those symbols, but this interface does not look consistent.

fbgemm_gpu/test/uvm_test.py
```diff
@@ -78,8 +81,12 @@ def test_uvm_to_cpu(self, sizes: List[int], vanilla: bool) -> None:
     @unittest.skipIf(*gpu_unavailable)
     def test_enum(self) -> None:
         # pyre-ignore[16]
-        assert cudaMemoryAdvise.cudaMemAdviseSetAccessedBy.value == 5
+        if torch.version.hip:
+            assert hipMemoryAdvise.hipMemAdviseSetAccessedBy.value == 5
+        else:
+            assert cudaMemoryAdvise.cudaMemAdviseSetAccessedBy.value == 5
 
+    @skipIfRocm
```
shintaro-iwasaki, line 89: Why are all these unit tests skipped?

fbgemm_gpu/bench/quantize_ops_benchmark.py
liligwu: Quite sorry for this, I overlooked the conflict. It's resolved.

fbgemm_gpu/include/fbgemm_gpu/enum_utils.h
liligwu: This fix works on CUDA as well. Unit tests in `uvm_test.py` passed on CUDA.

fbgemm_gpu/test/uvm_test.py
liligwu: Changes on hipify_torch needed for this issue.

fbgemm_gpu/test/uvm_test.py
liligwu: It complains "out of memory" on some of the ROCm devices. We're investigating it.

amathews-amd: Issue is because a large amount of memory (>1TB) was allocated in the tests. We are checking if this is valid.

fbgemm_gpu/bench/quantize_ops_benchmark.py
shintaro-iwasaki: Thanks for your quick fix!

fbgemm_gpu/test/uvm_test.py
shintaro-iwasaki: I truly understand why it happens and this is the easy fix, but I don't think this is the right interface we should provide since PyTorch and FBGEMM basically provides GPU features as "CUDA" if the corresponding features are provided by HIP/ROCm. Can you escape `hipify_python.py` somehow?

shintaro-iwasaki: Any updates on this @liligwu ?

jithunnair-amd: @shintaro-iwasaki @jianyuh I think this issue is cropping up because of the (unnecessary?) clash in the naming of the uvm enum `cudaMemoryAdvise` (https://github.com/pytorch/FBGEMM/blob/a70d14ca5f7f3354337d3ccfa5558c81fa550a00/fbgemm_gpu/src/cumem_utils.cu#L360) vs the CUDA-provided enum `cudaMemoryAdvise` (which has a HIP equivalent `hipMemoryAdvise` that needs to be used for ROCm). Are they required to be identically named? If not, I think the most straightforward solution here is to deduplicate the naming.

I agree that the python interface should be unified and hence the python enum should be the same for ROCm vs CUDA. That will also render changes like these unnecessary:
https://github.com/pytorch/FBGEMM/pull/1142/files#diff-063346a73da0b0a6d9cb2ba902a1635ae9f57beeda2df107648bf4724cf47962R85

shintaro-iwasaki: Thanks for your comment @jithunnair-amd. This PR provides the following for HIP:
```
from fbgemm_gpu.uvm import cudaMemAdvise, hipMemoryAdvise, cudaMemPrefetchAsync
```
This is not a right interface; the names are very inconsistent.


There should be four directions to implement this.

1. HIP provides both "hip" API and "cuda" alias
   - HIP provides both "hipMemAdvise", "hipMemoryAdvise", "hipMemPrefetchAsync", "cudaMemAdvise", "cudaMemoryAdvise", "cudaMemPrefetchAsync"
   - CUDA provides "cudaMemAdvise", "cudaMemoryAdvise", "cudaMemPrefetchAsync"

This sounds great to me if this is the proposal.

2. HIP provides "cuda"
   - HIP provides: "cudaMemAdvise", "cudaMemoryAdvise", "cudaMemPrefetchAsync"
   - CUDA provides "cudaMemAdvise", "cudaMemoryAdvise", "cudaMemPrefetchAsync"

This is what I proposed; if the hipification mechanism is somehow escaped, it should be easy.

3. HIP provides "hip" only
   - HIP provides: "hipMemAdvise", "hipMemoryAdvise", "hipMemPrefetchAsync"
   - CUDA provides "cudaMemAdvise", "cudaMemoryAdvise", "cudaMemPrefetchAsync"


We don't want this since the interface is not unified. We would appreciate option 1 for compatibility.

4. HIP mixes "hip" and "cuda"
   - HIP provides: "cudaMemAdvise", "hipMemoryAdvise", "cudaMemPrefetchAsync"
   - CUDA provides "cudaMemAdvise", "cudaMemoryAdvise", "cudaMemPrefetchAsync"

This is the current implementation. This interface is messy

---

I might be misunderstanding your point. Would you mind teaching us your proposal, @jithunnair-amd ?

jithunnair-amd: @shintaro-iwasaki Sure. What you want here is the separation of "church" (Python) and "state" (C++). In other words, the Python enum need not be exactly same as the C++ enum, however, IIUC, in this case, you're forcing that via the following code: https://github.com/pytorch/FBGEMM/blob/a70d14ca5f7f3354337d3ccfa5558c81fa550a00/fbgemm_gpu/fbgemm_gpu/uvm.py#L21

This creates a problem for hipify because you ideally want hipify to only modify the C++-side uses of CUDA APIs/enums, but *not* the python side. However, since you use the above code in conjunction with the following way to define the uvm enum (which then gets mimicked on the python side using the above code):
https://github.com/pytorch/FBGEMM/blob/a70d14ca5f7f3354337d3ccfa5558c81fa550a00/fbgemm_gpu/src/cumem_utils.cu#L361

Hipify will replace the identifier it thinks is a CUDA identifier here, because it doesn't know that it'll also be eventually used as the name of the enum on the Python side. That is why you end up having it show up here:
https://github.com/pytorch/FBGEMM/pull/1142/files#diff-063346a73da0b0a6d9cb2ba902a1635ae9f57beeda2df107648bf4724cf47962R27

**PROPOSAL**
Our proposal is the following:
1. Rename the python enums to be distinct from the CUDA C types eg.
`from fbgemm_gpu.uvm import gpuMemAdvise, gpuMemoryAdvise, gpuMemPrefetchAsync`
2. Use a different macro to map the python/uvm enum name to the CUDA type:
```
+#define FBGEMM_GPU_ENUM_ITEM2(x, y) \
+ { #x, y }
```
```
FBGEMM_GPU_ENUM_REGISTER_START(uvm, gpuMemoryAdvise){
 FBGEMM_GPU_ENUM_ITEM2(gpuMemAdviseSetReadMostly, cudaMemAdviseSetReadMostly),
```
3. This will make the python side quite clean as well:
```
assert gpuMemoryAdvise.gpuMemAdviseSetAccessedBy.value == 5
...
cudaMemAdvise(uvm_t, gpuMemoryAdvise.gpuMemAdviseSetAccessedBy)
```

This gives us a unified interface in python without hacking the hipify paradigm.

shintaro-iwasaki: We really want to keep the existing interface. It's really great if we can add another name (`gpuMemXXX`), but please keep the existing Python interface.

It's a hipification script issue, and I believe we can easily escape it.  For example, token concat (`##`) can avoid hipification without changing `hipify_python.py`. There should be a reasonably elegant way to keep the interface. To do so, the following you suggested would be a good idea.
```
#define FBGEMM_GPU_ENUM_ITEM2(x, y) \
 { #x, y }
```

liligwu: Hi @shintaro-iwasaki , we pushed a workaround for this Python API issue. Please review.

shintaro-iwasaki: Thanks. Not very clean (understandable though), but it looks good to me!

fbgemm_gpu/test/uvm_test.py
shintaro-iwasaki: Is there any progress here?

fbgemm_gpu/bench/quantize_ops_benchmark.py
```diff
@@ -138,6 +128,40 @@ def bench(
...
+    num_columns: int,
+    num_rows: int,
+    warmup_runs: int,
+) -> None:
+    if num_columns == -1 or num_rows == -1:
+        bench_spectrum(flush_gpu_cache_size_mb=flush_gpu_cache_size_mb, iters=iters, warmup_runs=warmup_runs)
+    else:
+        bench_impl(flush_gpu_cache_size_mb=flush_gpu_cache_size_mb, iters=iters, num_columns=num_columns, num_rows=num_rows, warmup_runs=warmup_runs)
+
+
```
shintaro-iwasaki, line 164: Sorry, I missed this in the first review pass. I might be just missing the context, but why do we need to update the quantization test?

fbgemm_gpu/test/uvm_test.py
liligwu: Sorry, not yet. I've submitted a JIRA ticket internally, but it may take some time.

fbgemm_gpu/bench/quantize_ops_benchmark.py
amathews-amd: Our discussions require us testing specific versions of `num_rows` and `num_columns`, not in the standard spectrum tested here. The old implementation of this file had `num_rows` and `num_columns` as arguments. This is bringing some of that functionality back.

liligwu:
Hi @shintaro-iwasaki , Since all of your comments were addressed. Would you please import the changes of this PR, so ROCm will be enabled? Thank you.

fbgemm_gpu/test/uvm_test.py
shintaro-iwasaki: I got it. Thanks for your updates.

shintaro-iwasaki:
Thanks @liligwu. Please let us work on internal testing.

shintaro-iwasaki:
(Fixing internal issues. Please give us a few more days.)

liligwu:
> 

Take your time. Thank you for letting us know.

PR #1161, opened 2022-06-17
Enable ROCm build-only CI

liligwu:
Enable ROCm build-only CI


amathews-amd:
cc: @shintaro-iwasaki

.github/workflows/fbgemmci.yml
```diff
@@ -248,6 +248,67 @@ jobs:
...
 
+  build_amd_gpu:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+
+    steps:
+    - name: Free space
+      run: sudo rm -rf /usr/local/android /usr/share/dotnet /usr/local/share/boost /opt/ghc /usr/local/share/chrom* /usr/share/swift /usr/local/julia* /usr/local/lib/android
```
shintaro-iwasaki, line 259: I don't think we need this step.

.github/workflows/fbgemmci.yml
```diff
@@ -248,6 +248,67 @@ jobs:
...
+
+    - uses: actions/checkout@v2
+
+    - name: Install ROCm
```
shintaro-iwasaki, line 263: nit: could you name it ROCm 5.1.1 (or a certain ROCm version used in this CI)? We want to use the same name as that of CUDA.

fbgemm_gpu/cmake/Hip.cmake
```diff
@@ -106,22 +106,132 @@ find_package(HIP)
...
+      message(FATAL_ERROR "Caffe2: Couldn't determine version from header: " ${output_var})
+    endif()
+    message(STATUS "Caffe2: Header version is: " ${rocm_version_from_header})
+    set(ROCM_VERSION_DEV_RAW ${rocm_version_from_header})
+    message("\n***** ROCm version from rocm_version.h ****\n")
+
+  # ROCM < 4.5, we don't have the header api file, use flat file
+  else()
+    file(READ "${ROCM_PATH}/.info/version-dev" ROCM_VERSION_DEV_RAW)
+    message("\n***** ROCm version from ${ROCM_PATH}/.info/version-dev ****\n")
```
shintaro-iwasaki, line 146: I don't think the current FBGEMM supports ROCm < 4.5.

https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/src/merge_pooled_embeddings_gpu.cpp#L99 uses `rsmi_is_P2P_accessible()`, and this API was added since ROCm 5.0.0: https://github.com/RadeonOpenCompute/rocm_smi_lib/releases

fbgemm_gpu/cmake/Hip.cmake
```diff
@@ -106,22 +106,132 @@ find_package(HIP)
...
+  execute_process(COMMAND dpkg -l COMMAND grep rocm-libs COMMAND awk "{print $2 \" VERSION: \" $3}")
+  execute_process(COMMAND dpkg -l COMMAND grep hsakmt-roct COMMAND awk "{print $2 \" VERSION: \" $3}")
+  execute_process(COMMAND dpkg -l COMMAND grep rocr-dev COMMAND awk "{print $2 \" VERSION: \" $3}")
+  execute_process(COMMAND dpkg -l COMMAND grep -w hcc COMMAND awk "{print $2 \" VERSION: \" $3}")
+  execute_process(COMMAND dpkg -l COMMAND grep hip-base COMMAND awk "{print $2 \" VERSION: \" $3}")
+  execute_process(COMMAND dpkg -l COMMAND grep hip_hcc COMMAND awk "{print $2 \" VERSION: \" $3}")
+
+  message("\n***** Library versions from cmake find_package *****\n")
+
+  if(ROCM_VERSION_DEV VERSION_GREATER_EQUAL "5.1.0")
```
shintaro-iwasaki, line 181: Just out of curiosity, do you need an additional branch for `5.1.0` as well as `5.0.0`?

fbgemm_gpu/CMakeLists.txt
```diff
@@ -48,8 +48,10 @@ if(DEFINED GLIBCXX_USE_CXX11_ABI)
   if(${GLIBCXX_USE_CXX11_ABI} EQUAL 1)
     set(CXX_STANDARD_REQUIRED ON)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=1")
-    message("${CMAKE_CXX_FLAGS}")
+  else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=0")
   endif()
+  message("${CMAKE_CXX_FLAGS}")
```
shintaro-iwasaki, line 54: Nice catch. Thanks.

.github/workflows/fbgemmci.yml
```diff
@@ -248,6 +248,67 @@ jobs:
...
+
+    - name: Test fbgemm_gpu installation
+      shell: bash
+      run: |
+        cd fbgemm_gpu
+        cd test
+        python3 input_combine_test.py
+        python3 quantize_ops_test.py
+        # disable sparse_ops_test for ROCm at the moment, since a "core dumped" aborts the test on the Github actions vm that has no GPUs.
+        # python3 sparse_ops_test.py
```
shintaro-iwasaki, line 308: We don't use AMD GPUs in this test configuration, but why does `sparse_ops_test` fail?

shintaro-iwasaki (commented):
Thanks for creating a PR! Overall it looks good to me, but please let me leave some comments.

.github/workflows/fbgemmci.yml
liligwu: Github actions VM is space-limited for ROCm and Pytorch. The CI throws a "No Space Left on Device" error when installing Pytorch.

.github/workflows/fbgemmci.yml
liligwu: > 

Sure, I'll add the version in the next commit. Thank you for reminding me.

.github/workflows/fbgemmci.yml
liligwu: Yeap, no GPUs in this config. We're investigating the failure. The test is passed when GPUs are available.

fbgemm_gpu/cmake/Hip.cmake
liligwu: Thank you for letting me know that. I'll discuss this with my folks and see if we will remove this branch.

jithunnair-amd: @liligwu I fully agree. Let's get rid of the `else` clause, since `rocm_version.h` is the way to get the ROCm version going forward.

fbgemm_gpu/cmake/Hip.cmake
jithunnair-amd: @shintaro-iwasaki I think this is reflecting a change in the structure of cmake files in ROCm installation from 5.0 to 5.1. That being said, if we don't need backward compatibility with ROCm5.0, we could get rid of the `else` branch.

shintaro-iwasaki: Thanks. We'd like to support more ROCm versions, but if only 5.1.0+ is supported, that's also fine. Could you please add a version check in CMakefile?

shintaro-iwasaki:
@liligwu Thanks for updating it! Please address the following.
- `sparse_ops_test` failure: why
- ROCm version check: ROCm >=5.0 or >=5.1?
   - Please print an error message if the version requirement is not satisfied.

The other changes look good to me.

.github/workflows/fbgemmci.yml
shintaro-iwasaki: Thanks! Just in case: if you keep this line commented out, please remove the trailing whitespace.
```py
        # python3 sparse_ops_test.py" " <- remove the white space here
```

liligwu:
> @liligwu Thanks for updating it! Please address the following.
> 
> * `sparse_ops_test` failure: why
> * ROCm version check: ROCm >=5.0 or >=5.1?
>   
>   * Please print an error message if the version requirement is not satisfied.
> 
> The other changes look good to me.

- `sparse_ops_test` failure is caused by `hipGetDeviceCount()` returns an improper error message since GPUs do not present. We have filed a JIRA ticket internally to HIP team.

- We decide to anchor on ROCm5.1.1, and an error message will be thrown if ROCm version number is less than 5.1.1.

The check for ROCm build passed. @shintaro-iwasaki , would you please close this PR if this build succeeds? Thank you.

PR #1194, opened 2022-07-08
Add pooling mode to device bench

zhuzilin:
This would help us benchmark EmbeddingCollection in torchrec.

Thank you for your time in reviewing this PR :)

### DIFF
--- a/fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
+++ b/fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
@@ -215,6 +215,17 @@ def device(  # noqa C901
 
     nparams = sum(w.numel() for w in emb.split_embedding_weights())
     param_size_multiplier = weights_precision.bit_rate() / 8.0
+    output_size_multiplier = output_dtype.bit_rate() / 8.0
+    if do_pooling:
+        read_write_bytes = (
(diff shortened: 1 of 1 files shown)


jianyuh:
Thanks! Just curious in your use case, do you have both unpooled embedding (none) and pooled embedding (sum), or only unpooled embedding (none)?

fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py
```diff
@@ -191,6 +205,7 @@ def device(  # noqa C901
             weights_precision=weights_precision,
             stochastic_rounding=stoc,
             output_dtype=output_dtype,
+            pooling_mode=pooling_mode,
```
jianyuh, line 208: Could you update the BW calculation formula in Line 254 and Line 279 ? Basically referring to https://github.com/pytorch/FBGEMM/blob/b3e129c7ab6486fd971dfbc62c443706c1d5ada3/fbgemm_gpu/bench/split_table_batched_embeddings_benchmark.py#L983-L991 . The number of bytes for unpooled embedding are increased by a factor of L.

zhuzilin: fixed.

jianyuh (approved):

PR #1197, opened 2022-07-09
Add nobag kernel for embedding_dim <= 32

zhuzilin:
This is an attempt to optimize the nobag forward kernel for tables whose embedding dim is smaller or equal to 32. I was exploring this as some of our production models have embedding_dim = 32. The optimization results in 10%~30% enhancement for small embedding_dim and could be applied to other kernels. However, it's worth noticing that a 10% enhancement on 1 kernel can barely have any effect on the overall training speed. Therefore, I'm totally fine with whether this optimization gets accepted, just trying to share some ideas we had to prevent others' repetitive work :)

The main rationale is that the current implementation will use all 32 threads in a warp to load 1 embedding vector, which means when the embedding dim is smaller than 128, some threads in the warp do nothing but wait. This PR will split threads into groups, e.g. for embedding_dim=32,  each group has 8 threads, and let the threads process each embedding vector in group instead of in warp.

The performance enhancement of this trick is benchmarked with https://github.com/pytorch/FBGEMM/pull/1194:
```
python bench/split_table_batched_embeddings_benchmark.py device --pooling=none --iters=1000 --embedding-dim=$EMB_DIM
```
And figures are:
|embedding dim|4|8|16|32|
|---------------|-|-|-|-|
|before(us)|136|138|142|154|
|after(us)|96|100|113|141|

Thank you for your time on this PR and it will be great if you could share your thoughts on this type of optimization :)

### DIFF
--- a/fbgemm_gpu/codegen/embedding_forward_split_template.cu
+++ b/fbgemm_gpu/codegen/embedding_forward_split_template.cu
@@ -26,6 +26,126 @@ constexpr size_t kForwardMaxThreads = 512;
 using Tensor = at::Tensor;
 using namespace fbgemm_gpu;
 
+{% if not weighted %}
+template <
+    typename emb_t,
+    typename cache_t,
+    {% if not dense %}
+    typename output_t,
+    {% endif %}
+    typename index_t,
+    size_t kThreadGroupSize
+    >
+__launch_bounds__(kForwardMaxThreads)
+__global__ void {{ "dense" if dense else "split" }}_embedding_nobag_codegen_forward_unweighted_small_kernel(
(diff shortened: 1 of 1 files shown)


jianyuh:
Thanks for the great optimization!  10%~30% improvement for embedding dim = 32 case in the op level is impressive. Wonder if you can also help evaluate the increase of the binary size with this PR? Basically checking the generated fbgemm_gpu_*.so file binary size increase.

zhuzilin:
@jianyuh The binary size of `fbgemm_gpu_py.so` is 275873080 bytes before and 276987184 after, increasing by around 1M.

Issue #1201, opened 2022-07-11
Latter prefetch may evict previous prefetched embeddings


zhuzilin:
Our team is trying to utilize the `SplitTableBatchedEmbeddingBagsCodegen.prefetch` to accelerate training in torchrec. And during the survey, I found that the `lru_cache_insert_kernel` and `lfu_cache_insert_kernel` may evict the id prefetched last time. For example, in the following pipeline, if `prefetch_2`, which is prefetching the data for `compute_2`, may accidentally evict the data being used by `compute_1`.
```
stream 0              |    compute_0    |    compute_1    |    compute_2    | ...
stream 1 | prefetch_0 | prefetch_1 |    | prefetch_2 |    | prefetch_3 | ...
```
A quick fix for LRU is passing the first element of `self.timesteps_prefetched` as the eviction threshold in `lru_cache_insert` kernels. But I haven't figured out a good way to fix LFU kernels...

Also, a minor issue, the current `benchmark_pipelined_requests` doesn't seem to do pipelining of `prefetch` and `forward`, as they are using the same stream.

colin2328:
cc @jianyuh @jspark1105

jianyuh:
Thanks for reporting the issue! Feel free to submit a PR on fixing LRU and `benchmark_pipelined_requests` benchmark (https://github.com/pytorch/FBGEMM/blob/c1a63cb6b9e3f369619ecd53f7aadf1a6d9384f3/fbgemm_gpu/bench/bench_utils.py#L385 with multiple streams). For LFU, not sure if it is easy to create some unique tag for the frequency counter.

zhuzilin:
@jianyuh 
> For LFU, not sure if it is easy to create some unique tag for the frequency counter.

I was also thinking about using maybe a bit as tag for "being used". However, it's hard to determine when could we untag them, as user may or may not run backward on the model...

> Feel free to submit a PR

I'll try for a PR :) (may be 2 separate ones for them)

jasonjk-park:
SW cache is not thread-safe (or multi-stream safe in this case) - prefetching and computing at the same time is not the intention of the original design. Having some sort of lock/mutex would make it work, however, that'd prevent us to use in pipelined fashion as well.

It'd be better to support multi-stream safeness completely (not just LRU, but also LFU, and probably replacement policy agnostic way) rather than giving false impression that it works by partially allowing it.

zhuzilin:
@jasonjk-park @jianyuh 
[...]

PR #1203, opened 2022-07-12
Fix LRU cache prefetch

zhuzilin:
Fix issues in #1201.

I'm not sure `non_evictable_time_stamp` is a suitable name :p

Thank you for your time on this PR.

cc @jianyuh

### DIFF
--- a/fbgemm_gpu/src/split_table_batched_embeddings.cpp
+++ b/fbgemm_gpu/src/split_table_batched_embeddings.cpp
@@ -35,7 +35,8 @@ void lru_cache_populate_cuda(
     Tensor lxu_cache_weights,
     int64_t time_stamp,
     Tensor lru_state,
-    bool stochastic_rounding);
+    bool stochastic_rounding,
+    int64_t non_evictable_time_stamp = -1);
 
 // LRU cache: fetch the rows corresponding to `linear_cache_indices` from
 // `weights`, and insert them into the cache at timestep `time_stamp`.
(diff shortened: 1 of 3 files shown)


fbgemm_gpu/src/split_table_batched_embeddings.cpp
```diff
@@ -115,7 +116,7 @@ TORCH_LIBRARY_FRAGMENT(fb, m) {
       "linearize_cache_indices(Tensor cache_hash_size_cumsum, Tensor indices, Tensor offsets) -> Tensor");
   DISPATCH_TO_CUDA("linearize_cache_indices", linearize_cache_indices_cuda);
   m.def(
-      "lru_cache_populate(Tensor weights, Tensor hash_size_cumsum, int total_cache_hash_size, Tensor cache_index_table_map, Tensor weights_offsets, Tensor D_offsets, Tensor linear_cache_indices, Tensor(a!) lxu_cache_state, Tensor(b!) lxu_cache_weights, int time_stamp, Tensor(c!) lru_state, bool stochastic_rounding) -> ()");
+      "lru_cache_populate(Tensor weights, Tensor hash_size_cumsum, int total_cache_hash_size, Tensor cache_index_table_map, Tensor weights_offsets, Tensor D_offsets, Tensor linear_cache_indices, Tensor(a!) lxu_cache_state, Tensor(b!) lxu_cache_weights, int time_stamp, int non_evitable_time_stamp, Tensor(c!) lru_state, bool stochastic_rounding) -> ()");
```
jianyuh, line 119: Could you move `int non_evitable_time_stamp,` as the last argument and assign a default value? This is primarily for backward compatibility.

zhuzilin: fixed.

colin2328:
@zhuzilin  from @jasonjk-park  :
"looked at the original issue, but is it safe to do such pipelining between prefetch and compute? it's not thread-safe as cache is being used and updated at the same time. seems like this is a hacky way to get around that issue, but not entirely for any replacement policy?"

PR #1221, opened 2022-07-26
Add prototype for optimized dense2jagged kernel for 1 jagged dim

chirayuG-nvidia:
Sharing optimized prototype kernel for 2D dense_to_jagged operations using vectorized operations.
CC: @mjanderson09

### DIFF
--- a/fbgemm_gpu/src/jagged_tensor_ops.cu
+++ b/fbgemm_gpu/src/jagged_tensor_ops.cu
@@ -143,6 +143,381 @@ __launch_bounds__(kMaxThreads) void jagged_dense_elementwise_dense_output_kernel
   }
 }
 
+
+// brute-force offset generation kernel - not used, performance can be tuned
+// can be used to overcome shared memory size constraints
+template <typename index_t>
+__global__ void dense_offset_to_jagged_map(
+  index_t* dense_offsets,
+  index_t* dense_to_jagged_map,
+  int32_t num_offsets,
+  int32_t jagged_dim_stride,
+  int32_t e_dim_stride) {
+
+  int32_t idx = threadIdx.x + blockIdx.x * blockDim.x;
+  if (idx < (num_offsets-1)) {
+    index_t start_offset = dense_offsets[idx];
+    index_t nnz = dense_offsets[idx + 1] - start_offset;
+
+    index_t stride = (idx * jagged_dim_stride);
+
+    for (index_t k = 0; k < nnz; k++) {
+      dense_to_jagged_map[start_offset + k] = stride + (k * e_dim_stride);
+    }
+  }
+}
+
+template <typename index_t>
+__global__ void dense_offset_to_jagged_map_opt_kernel_(
+  index_t* dense_offsets,
+  index_t* dense_to_jagged_map,
+  int32_t num_offsets,
+  int32_t total_jagged_rows,
+  int32_t jagged_dim_stride,
+  int32_t e_dim_stride) {
+
+  extern __shared__ index_t smem_dense_offsets[];
+
+  // should just stream from L2 to all smem's
+  for (int32_t i = threadIdx.x; i < num_offsets; i += blockDim.x) {
+    smem_dense_offsets[i] = dense_offsets[i];
+  }
+  __syncthreads();
+
+  for (index_t idx = (threadIdx.x + blockIdx.x * blockDim.x); idx < total_jagged_rows;
+          idx += (blockDim.x * gridDim.x)) {
+    // search for idx in smem_dense_offsets
+    index_t mid = num_offsets / 2;
+    index_t lower = 0, upper = (num_offsets-1); // offset shoud have atleast 2 values
+    index_t res = 0;
+
+    while (lower <= upper) {
+      if (idx >= smem_dense_offsets[mid]) {
+        lower = mid + 1;
+        res = mid;
+      } else {
+        upper = mid - 1;
+      }
+      mid = (lower + upper) / 2;
+    }
+    // warp sync ?
+    index_t ix = idx - smem_dense_offsets[res];
+    index_t stride = (res * jagged_dim_stride); // res is row in dense-0th dim
+    dense_to_jagged_map[idx] = stride + (ix * e_dim_stride);
+  }
+}
+
+template <typename scalar_t>
+struct VecType128{};  // 128 bit load-store vector type
+
+template <>
+struct VecType128<c10::Half> {
+  typedef float4 TType; // Transaction Type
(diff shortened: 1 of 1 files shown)


fbgemm_gpu/src/jagged_tensor_ops.cu
```diff
@@ -143,6 +143,381 @@ __launch_bounds__(kMaxThreads) void jagged_dense_elementwise_dense_output_kernel
...
+        lower = mid + 1;
+        res = mid;
+      } else {
+        upper = mid - 1;
+      }
+      mid = (lower + upper) / 2;
+    }
+    // warp sync ?
+    index_t ix = idx - smem_dense_offsets[res];
+    index_t stride = (res * jagged_dim_stride); // res is row in dense-0th dim
```
jspark1105, line 205: Strides are manually used vs. tensor accessor. Does it make a difference in perf?

fbgemm_gpu/src/jagged_tensor_ops.cu
```diff
@@ -143,6 +143,381 @@ __launch_bounds__(kMaxThreads) void jagged_dense_elementwise_dense_output_kernel
...
+  extern __shared__ index_t smem_dense_offsets[];
+
+  // should just stream from L2 to all smem's
+  for (int32_t i = threadIdx.x; i < num_offsets; i += blockDim.x) {
+    smem_dense_offsets[i] = dense_offsets[i];
+  }
+  __syncthreads();
+
+  for (index_t idx = (threadIdx.x + blockIdx.x * blockDim.x); idx < total_jagged_rows;
+          idx += (blockDim.x * gridDim.x)) {
```
jspark1105, line 188: nit: we can use CUDA_KERNEL_LOOP(idx, total_jagged_rows) which is defined in pytorch/aten/src/ATen/cuda/detail/KernelUtils.h

fbgemm_gpu/src/jagged_tensor_ops.cu
```diff
@@ -143,6 +143,381 @@ __launch_bounds__(kMaxThreads) void jagged_dense_elementwise_dense_output_kernel
...
+) {
+  // for some reason out rows in jagged can be larger than in dense
+  // for now assume same as elements in dense_to_jagged_map
+
+  typedef typename VecType128<scalar_t>::TType vec128;
+  typedef typename VecType64<scalar_t>::TType vec64;
+  typedef typename VecType32<scalar_t>::TType vec32; //  no-op for double
+
+  // 128 bit alignment check
+  if (((e_dim * sizeof(scalar_t)) << 3 ) % 128 == 0) {
```
jspark1105, line 444: Instead of applying vectorization sort of all-or-nothing, can we do best effort like the first loop applied up to columns rounded down to 128-bit, the second loop applied to up to columns rounded down to 64-bit, and so on? In this way when e_dim is say 576-bit we can use 128-bit vectorization up to 512-bit and then 64-bit vectorization. Or this will regress perf?

jspark1105 (commented):
This is amazing! Thanks for the PR. I added a few comments here for visibility but these can be certainly addressed on our side.

fbgemm_gpu/src/jagged_tensor_ops.cu
chirayuG-nvidia: It shouldn't make any difference in perf as long as accessor is also adding same amount of instructions. I used explicit calculations as it makes the code self explanatory.

fbgemm_gpu/src/jagged_tensor_ops.cu
chirayuG-nvidia: Performance will depend on the overhead of math instructions get added to do the alignment check by each thread of the warp every time it executes.  I think you can find both types of problem size where doing the incremental vector-bit decrement will help performance and other (perfectly aligned cases) where it will reduce it.
IMO ideal solution is to allocate contiguous tensors with appropriate pitch in each dimension that aligns the vectorized load/store. Not sure if pytorch does that already.

mjanderson09: So you're only checking alignment with respect to e_dim, not alignment of the actual pointers, right? If that's the case, can we just do the following? Or do we have to check alignment of "values" and "dense" pointers. 

```
    VecType128::TType * values128 = (VecType128::TType *)&(values[0]);
    VecType128::TType * dense128 = (VecType128::TType *)&(dense[0]);
    VecType64::TType * values64 = (VecType64::TType *)&(values[0]);
    VecType64::TType * dense64 = (VecType64::TType *)&(dense[0]);
    VecType32::TType * values32 = (VecType32::TType *)&(values[0]);
    VecType32::TType * dense32 = (VecType32::TType *)&(dense[0]);
    for(int real_row = values_row ; real_row < nnz ; real_row += blockDim.y * gridDim.y) {
        int dense_row = rows[real_row];
        int dense_col = cols[real_row];
        for(int tid = threadIdx.x ; tid < E/8 ; tid += blockDim.x) {
            values128[tid + real_row * (E/8)] = dense128[tid + dense_col * (E/8) + dense_row * L * (E/8)];
        }
        for(int tid = threadIdx.x + (E/8)*8 ; tid < E/4 ; tid += blockDim.x) {
            values64[tid + real_row * (E/4)] = dense64[tid + dense_col * (E/4) + dense_row * L * (E/4)];
        }
        for(int tid = threadIdx.x + (E/4)*4 ; tid < E/2 ; tid += blockDim.x) {
            values32[tid + real_row * (E/2)] = dense32[tid + dense_col * (E/2) + dense_row * L * (E/2)];
        }
        for(int tid = threadIdx.x + (E/2)*2 ; tid < E ; tid += blockDim.x) {
            values[tid + real_row * (E)] = dense[tid + dense_col * (E) + dense_row * L * (E)];
        }
    }
```

chirayuG-nvidia: I am doing a alignment check with `e_dim * sizeof(scalar_t)` because :
1. It will ensure jagged_out tensor is aligned
2. Because of broadcast and the only effective stride in dense tensor being same as e_dim, so my check ensures overall things will align for all the consequent pointer location that  kernel will try to read from. 

The alternate that you suggested will run into pointer alignment error. For eg: say output jagged is 130 bit wide (65 half's), so very first row will work fine with your logic, however the 2nd row will try to write first 8 half's with address that is not aligned for 128-bit operation.

> for(int tid = threadIdx.x ; tid < E/8 ; tid += blockDim.x) {
Note that in my code I use a warp to service two rows (block.x = 16) , I use `tid = threadIdx.x % blockDim.x`

PR #1235, opened 2022-08-03
Print error msg when loading fbgemm_gpu_py.so

zhuzilin:
The old error message may be confusing when the error is not that `fbgemm_gpu_py.so` cannot be found, e.g. CPU version of pytorch is installed and `libc10_cuda.so` cannot be found. User could still be able to find out that `fbgemm_gpu_py.so` is not there, as the error message now is:
```
/path/to/fbgemm_gpu/fbgemm_gpu_py.so: cannot open shared object file: No such file or directory
```

Thank you for your time on this PR:)


brad-mengchi:
Thanks @zhuzilin! This is helpful to show the exact reason for debugging.

PR #1237, opened 2022-08-04
Run CI on pytorch built with CUDA 11.3

mjanderson09:
Summary:
- pytorch has not updated their CUDA 11.5 .whl files with latest pytorch (1.13), while they have updated their CUDA 11.3 .whl files
- pytorch built with CUDA 11.3 will work with later CUDA versions, as CUDA is backward compatable
- Suggest we set a policy where we support building/installing FBGEMM with pytorch .whl which is built with CUDA 11.3 as this seems to be their preferred build. This guidance in our README and also torchrec README.
- This diff updates CI to install pytorch .whl which was built with CUDA 11.3, even when testing FBGEMM with later CUDA versions

Differential Revision: D38416122

### DIFF
--- a/.github/workflows/fbgemmci.yml
+++ b/.github/workflows/fbgemmci.yml
@@ -224,10 +224,8 @@ jobs:
         sudo apt-get update
         sudo apt-get -y install git pip python3-dev
         sudo pip install cmake scikit-build ninja jinja2 numpy hypothesis --no-input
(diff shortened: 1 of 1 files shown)


shintaro-iwasaki (approved):
We should use a truly "nightly" PyTorch build.
I overlooked this. Thanks!

PR #1240, opened 2022-08-05
Change the hardcoded 128 items per warp in embeddingBag to a variable and optimize for ROCm

liligwu:
Make @weihanmines's PR https://github.com/ROCmSoftwarePlatform/FBGEMM/pull/13 upstreamable. 
@sryap, would you please review the PR and consider converting it to a draft? Thank you.


jianyuh:
Do you have the perf improvement result on AMD GPUs with this PR? cc @shintaro-iwasaki @sryap @MohammadMahdiJavanmard

shintaro-iwasaki (commented):
As @jianyuh said, it'd be great if AMD could give us some performance numbers here.
If there's a concern on sharing performance numbers publicly, please send us privately.

Overall, the change looks good to me. I will import this PR and test it internally.

fbgemm_gpu/codegen/embedding_backward_code_generator.py
```diff
@@ -7,6 +7,7 @@
 
 import argparse
 import os
+import torch
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Tuple, Union
```
shintaro-iwasaki, line 13: Can I ask you to move `import torch` to address an internal lint error?

```py
# Current
import os
import torch
from dataclasses import dataclass
from typing import Any, Dict, List, Optional, Tuple, Union

import jinja2
```

```py
# New
import os
from dataclasses import dataclass
from typing import Any, Dict, List, Optional, Tuple, Union

import jinja2
import torch
```

sryap:
@jianyuh We see performance improvement in some cases (mostly the backward TBE) but we have requested AMD to share a more in-depth analysis with us (https://github.com/ROCmSoftwarePlatform/FBGEMM/pull/13#issuecomment-1205815995) to understand why this would improve the performance.

We think that the performance improvement comes from the reduced number of registers used per thread block, but we do not have a concrete evidence to show this at the moment.

amathews-amd:
> Do you have the perf improvement result on AMD GPUs with this PR? cc @shintaro-iwasaki @sryap @MohammadMahdiJavanmard

Please see slide 5 of 20220801_Mathews_AMDPhantom_MI250_Perf_EmbBag_DCPPv2 in workplace.

fbgemm_gpu/codegen/embedding_backward_code_generator.py
```diff
@@ -11,6 +11,7 @@
...
 
 import jinja2
+import torch
```
shintaro-iwasaki, line 14: Sorry, another thing. Can we make this script PyTorch independent (i.e., remove `import torch`)? We do not always run this code generator after fully installing PyTorch. It'd be nice if this script can take an argument from CMake when it runs.

shintaro-iwasaki: Perhaps here you can pass another ROCm-specific argument:
https://github.com/pytorch/FBGEMM/blob/5ce897d4bd46a59aac5f93459239c206f20ef15f/fbgemm_gpu/CMakeLists.txt#L174-L175

shintaro-iwasaki: Or an internal comment from @ajtulloch This should be easier to implement and maintain.

> Can’t you just do 4 * kWarpSize (or C10_WARP_SIZE or w/e the compile time warp size constant is) and avoid having to put this logic in the template system?

shintaro-iwasaki:
Hi @liligwu Thanks for your update. If we could use `4 * kWarpSize` (see above), that would be easier to maintain this parameter for both AMD and NVIDIA GPUs. I'd appreciate it if you could take a look at this option.

liligwu:
> Hi @liligwu Thanks for your update. If we could use `4 * kWarpSize` (see above), that would be easier to maintain this parameter for both AMD and NVIDIA GPUs. I'd appreciate it if you could take a look at this option.

Hi @shintaro-iwasaki , `4 * kWarpSize` is a good idea, but `kWarpSize` is a C++ constant, and we are using these `128/256` values in  `{ }`, which are in jinja template sections. Simply replacing `items_per_warp` with  `4 * kWarpSize` will trigger the following error:
> jinja2.exceptions.UndefinedError: 'kWarpSize' is undefined. 
I don't think we can "avoid having to put this logic in the template system" if it's in `{ }`. Or, we need to get the warp size in Python level.

jianyuh:
> > Hi @liligwu Thanks for your update. If we could use `4 * kWarpSize` (see above), that would be easier to maintain this parameter for both AMD and NVIDIA GPUs. I'd appreciate it if you could take a look at this option.
> 
> Hi @shintaro-iwasaki , `4 * kWarpSize` is a good idea, but `kWarpSize` is a C++ constant, and we are using these `128/256` values in `{ }`, which are in jinja template sections. Simply replacing `items_per_warp` with `4 * kWarpSize` will trigger the following error:
> 
> > jinja2.exceptions.UndefinedError: 'kWarpSize' is undefined.
> 
> I don't think we can "avoid having to put this logic in the template system" if it's in `{ }`. Or, we need to get the warp size in Python level.

We can use kWarpSize outside `{}`. I see kWarpSize is already used in this file (although it's defined in https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh#L32 ). In the future we should refactor and reuse the kWarpSize from c10.

liligwu:
> > > Hi @liligwu Thanks for your update. If we could use `4 * kWarpSize` (see above), that would be easier to maintain this parameter for both AMD and NVIDIA GPUs. I'd appreciate it if you could take a look at this option.
> > 
> > 
> > Hi @shintaro-iwasaki , `4 * kWarpSize` is a good idea, but `kWarpSize` is a C++ constant, and we are using these `128/256` values in `{ }`, which are in jinja template sections. Simply replacing `items_per_warp` with `4 * kWarpSize` will trigger the following error:
> > > jinja2.exceptions.UndefinedError: 'kWarpSize' is undefined.
> > 
> > 
> > I don't think we can "avoid having to put this logic in the template system" if it's in `{ }`. Or, we need to get the warp size in Python level.
> 
> We can use kWarpSize outside `{}`. I see kWarpSize is already used in this file (although it's defined in https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh#L32 ). In the future we should refactor and reuse the kWarpSize from c10.

Sorry to reply to you late. We are still working on this suggestion but didn't find a good solution for it, curretly; especially, the `items_per_wap` is in a jinja `for` loop, https://github.com/pytorch/FBGEMM/blob/86bf2ea27108da079b0b43e80acd0bc5ff5480a9/fbgemm_gpu/codegen/embedding_backward_split_indice_weights_template.cu#L233
though we can easily get `items_per_wap` in the `if` statements like https://github.com/pytorch/FBGEMM/blob/86bf2ea27108da079b0b43e80acd0bc5ff5480a9/fbgemm_gpu/codegen/embedding_backward_split_indice_weights_template.cu#L234 out of `{ }`

shintaro-iwasaki:
Thanks @liligwu for your investigation! We can revisit this issue on refactoring.  Would you mind rebasing this PR and making sure all tests pass?

shintaro-iwasaki (approved):

liligwu:
> Thanks @liligwu for your investigation! We can revisit this issue on refactoring. Would you mind rebasing this PR and making sure all tests pass?

I'll update the branch soon. Thank you.

shintaro-iwasaki:
@liligwu Merged. Thanks for your contribution!

liligwu:
> @liligwu Merged. Thanks for your contribution!

@shintaro-iwasaki , I appreciate your help.

PR #1261, opened 2022-08-26
fix index overflow for FindMinMax and QuantizeAvx2

XiaobingSuper:
There will have a data overflow for int type when input len > 2,147,483,647, PyTorch side has reported an issue which using FBGEMM Quantize, see https://github.com/pytorch/pytorch/issues/80501, the user example's input len is 5,117,410,688,  but FBGEMM side use **int** to represent the input len, there will get a wrong number for single thread case. 

This PR just fixed those two **Quantize** and **FindMinMax** API which are used in PyTorch side, there may have other functions which need to be updated using high-precison dtype.

### DIFF
--- a/include/fbgemm/QuantUtilsAvx2.h
+++ b/include/fbgemm/QuantUtilsAvx2.h
@@ -44,7 +44,7 @@ template <typename T = std::uint8_t, bool LEGACY = true>
 void QuantizeAvx2(
     const float* src,
     T* dst,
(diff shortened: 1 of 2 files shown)


XiaobingSuper:
@jianyuh

XiaobingSuper:
@jspark1105, could you help see the internal failed cases? I don't have access to see it. Thanks!

PR #1368, opened 2022-10-01
hip TBE Forward Kernel

dllehr-amd:
Table Batched Embedding Kernel for AMD GPUs.

This forward kernel takes advantage of GCN intrinsics to enable data pipelining for embedding bag indices loads while performing accumulations and stores.


amathews-amd:
cc: @shintaro-iwasaki @sryap

liligwu:
Mixed-dimension embeddingBag tables are enabled in the most recent commit. "test_cache_pipeline" is re-enabled accordingly.

PR #1412, opened 2022-10-21
Fix failure in AMD GPU test in test_reset_embedding_weight_momentum

doehyun:
Summary:
test_reset_embedding_weight_momentum fails only in AMD test (was ok in NVidia GPU).
It's added in D39046516 (https://github.com/pytorch/FBGEMM/commit/da2065d2e5a3032b38332c9747c5e4d8381a2453), but somehowe qparam_new is used without init, and
that shows different behavior between AMD and NVidia.

This diff initializes it to {1.0, and 0.0}, and that fixes AMD test failure.

Differential Revision: D40602633

### DIFF
--- a/fbgemm_gpu/src/split_embeddings_cache_cuda.cu
+++ b/fbgemm_gpu/src/split_embeddings_cache_cuda.cu
@@ -2553,7 +2553,7 @@ __global__ __launch_bounds__(kMaxThreads) void reset_weight_momentum_kernel(
     }
 
     // reset weight
-    float2 qparams_new;
+    float2 qparams_new = {1.0, 0.0}; // scaler=1.0, and offset=0.0, for int8.
     Vec4T<at::acc_type<cache_t, true>> weight_new; // 0 weight
     weight_row_template.store(
         weight_new,


liligwu:
**Analysis:**
1. The absence of initialization of float2 new_parames produces NAN on https://github.com/pytorch/FBGEMM/blob/a811ef6236c5b402a1cb6c40e7bdbc673108d3d7/fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh#L1033 for both CUDA and ROCm.
2. The NANs will further cause the inputs of `lrintf()` in https://github.com/pytorch/FBGEMM/blob/a811ef6236c5b402a1cb6c40e7bdbc673108d3d7/fbgemm_gpu/include/fbgemm_gpu/fbgemm_cuda_utils.cuh#L1034 be NANs. According to CUDA document https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1g639a876a55da8142dcd917ce6c12c27d the behavior of this NAN input is undefined. Although ROCm does not mention the outside-range behavior https://rocmdocs.amd.com/en/latest/Archived_Documentation/HIP-GUIDE.html?highlight=lrintf.

**Conclusions:** 
1. This is not a ROCm bug.
2. ROCm and CUDA handle the NANs differently when writing back.
3. `float2 qparams_new` should be initialized to avoid undefined behaviors in both platforms.

PR #1548, opened 2023-01-17
[CI][Test] Reenable TBE unit tests

shintaro-iwasaki:
The TBE unit test on GPUs is disabled because of its sporadic failure. This PR temporarily removes problematic configuration and reenables the TBE unit test for better test coverage


shintaro-iwasaki:
Please let me run the reenable test some times on GitHub Actions to check if it really passes.

shintaro-iwasaki:
We don't need this Diff as @q10 dealt with these cases nicely in his recent changes! Let me close this stale PR.

PR #1558, opened 2023-01-24
[build] Fix the nightly build

q10:
- Update the PyTorch installation to use `pip` instead of `conda` to work around the recent PyTorch nightlies missing the generated `torch/include/c10/cuda/impl/cuda_cmake_macros.h`

### DIFF
--- a/.github/workflows/fbgemm_nightly_build.yml
+++ b/.github/workflows/fbgemm_nightly_build.yml
@@ -70,8 +70,9 @@ jobs:
     - name: Install PyTorch and CUDA
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


q10:
Closing the PR as the original issue resolved itself.  To summarize, the last few nightly builds referenced the CPU-only version of PyTorch, which does not come with the necessary headers for compilation of GPU code:

```sh
pytorch-2.0.0.dev20230124  |      py3.7_cpu_0        85.4 MB  pytorch-nightly
```

vs

```sh
pytorch-2.0.0.dev20230124  |py3.7_cuda11.7_cudnn8.5.0_0        1.18 GB  pytorch-nightly
```

Also, for building, it is not recommended to install PyTorch and `nvcc` separately via conda and pip, respectively, it may come with unknown issues.

PR #1615, opened 2023-02-25
improve validation for input tensors to guard for case where inputs coming from different device

RoshanPAN:
Reviewed By: jianyuh, houseroad

Differential Revision: D43564925


RoshanPAN:
@jianyuh do we need you to help merge the PR?

PR #1649, opened 2023-03-16
Refactors for embedding_common

qxy11:
Summary: Converts if statements into a switch on the SparseType enum and adds more const into method param declarations

Differential Revision: D44140375


houseroad (approved):
LGTM

PR #1703, opened 2023-04-14
Fix exception in QuantUtilsTest

rlanday:
This test mistakenly calls reserve() to set a vector's length instead of resize(). reserve() allocates memory for the specified number of elements, but does not actually increase the number of elements that can legally be stored in the vector. This test runs with ASAN enabled which is catching this illegal access and causing the test to fail.

This change fixes the code to instead call resize(); the test now passes.


q10:
@rlanday Thanks for making the code changes!  The code will be merged shortly.

PR #1672, opened 2023-03-28
Move radix sort to common utilities and add the possibility to handle negative integers

DamianSzwichtenberg:
Summary:
Move the `radix_sort` implementation to common utilities, so it can be used in PyTorch in case it was not built with FBGEMM GPU.
Add the possibility to handle negative integers, which is crucial for reusing `radix_sort` in PyTorch's `sort` operation.

Details:
This PR addresses two issues:
1.  `radix_sort` is currently used in [scatter_reduce](https://github.com/dszwicht/pytorch/blob/master/aten/src/ATen/native/cpu/ScatterGatherKernel.cpp#L630) (please view this [comment](https://github.com/pytorch/pytorch/pull/82703/files#r1045360609) for more information). Till now `radix_sort` was under `fbgemm_gpu` subproject. It means that implementation was not available in PyTorch in case it was built for CPU - that's why `radix_sort` was copy pasted under aten directory in PyTorch. This PR moves `radix_sort` implementation to common utilities.
2. In GNN workloads we often sort 1D integer data with non-negative values, for example, when converting CSR to CSC format. Unfortunately, `torch.sort` for 1D data works sequentially. `radix_sort` seems to be a perfect match to accelerate described case. However, suppose we want to do that on the PyTorch site. In that case, we have to either fallback to a regular path after detecting negative numbers in the tensor or perform post-processing, by swapping positive and negative blocks of data (data like `[2, -1, -2, 1]` after sorting will be in the following form `[1, 2, -2, -1]`, due to the fact of how numbers are stored). Both these solutions are not elegant. As an alternative, I propose the extension of `radix_sort` algorithm, by giving it capability to work with negative numbers. This can be enabled by passing an optional parameter, `maybe_with_neg_vals`. If set to `true`, we will perform all passes (up to the most significant sign bit) and apply a special prefix sum combination in the last pass. An example of how we can reuse fbgemm in PyTorch can be found in my private fork, [here](https://github.com/dszwicht/pytorch/pull/2) (I also provide speedup data).

The above changes have several consequences:
1. `TORCH_CHECK` was replaced with `assert` as fbgemm CPU does not have PyTorch in its dependencies.
2. `__builtin_clz` was replaced with manual implementation as `__builtin_clz` is not portable.

Additional information for reviewers:
I did perform benchmarks of `radix_sort` before and after my code modification. I didn't observe any performance drop.

PR #1718, opened 2023-04-24
Fix data conversion in radix sort that can cause data loss

DamianSzwichtenberg:
Summary:
Fix data conversion in `radix_sort` that can cause data loss.

Details:
When `elements_count` is passed to the internal kernel implementation it is implicitly converted from `int64_t` to `int`. It can cause data loss, resulting in a partially sorted array. This PR fixes this issue. As a result of changing the `elements_count` type in internal functions to `int64_t`, `histogram` and `histogram_ps` types also were updated (to not generate further conversions).
This is a follow-up for #1672.

### DIFF
--- a/src/Utils.cc
+++ b/src/Utils.cc
@@ -598,12 +598,12 @@ size_t count_leading_zeros(T val) {
 constexpr int RDX_HIST_SIZE = 256;
 
 void update_prefsum_and_offset_in_range(
-    int& offset,
+    int64_t& offset,
     const int bins_beg,
     const int bins_end,
     const int nthreads,
-    const int* const histogram,
-    int* const histogram_ps) {
+    const int64_t* const histogram,
+    int64_t* const histogram_ps) {
   for (int bins = bins_beg; bins < bins_end; ++bins) {
     for (int t = 0; t < nthreads; ++t) {
       histogram_ps[t * RDX_HIST_SIZE + bins] = offset;
@@ -614,10 +614,10 @@ void update_prefsum_and_offset_in_range(
 
 void combine_prefix_sum(
     const int nthreads,
-    const int elements_count,
-    const int* const histogram,
-    int* const histogram_ps) {
-  int offset = 0;
+    const int64_t elements_count,
+    const int64_t* const histogram,
+    int64_t* const histogram_ps) {
+  int64_t offset = 0;
   update_prefsum_and_offset_in_range(
       offset, 0, RDX_HIST_SIZE, nthreads, histogram, histogram_ps);
   histogram_ps[RDX_HIST_SIZE * nthreads] = offset;
@@ -630,10 +630,10 @@ void combine_prefix_sum(
 
 void combine_prefix_sum_for_msb(
     const int nthreads,
-    const int elements_count,
-    const int* const histogram,
-    int* const histogram_ps) {
-  int offset = 0;
+    const int64_t elements_count,
+    const int64_t* const histogram,
+    int64_t* const histogram_ps) {
+  int64_t offset = 0;
   update_prefsum_and_offset_in_range(
       offset, 128, RDX_HIST_SIZE, nthreads, histogram, histogram_ps);
   update_prefsum_and_offset_in_range(
@@ -652,17 +652,17 @@ void radix_sort_kernel(
     const V* const input_values,
     K* const output_keys,
     V* const output_values,
-    const int elements_count,
-    int* const histogram,
-    int* const histogram_ps,
+    const int64_t elements_count,
+    int64_t* const histogram,
+    int64_t* const histogram_ps,
     const int pass,
     const bool pass_with_sign_bit = false) {
   const auto tid = omp_get_thread_num();
   const auto nthreads = omp_get_num_threads();
   const auto elements_count_4 = elements_count / 4 * 4;
 
-  int* const local_histogram = &histogram[RDX_HIST_SIZE * tid];
-  int* const local_histogram_ps = &histogram_ps[RDX_HIST_SIZE * tid];
+  auto* const local_histogram = &histogram[RDX_HIST_SIZE * tid];
+  auto* const local_histogram_ps = &histogram_ps[RDX_HIST_SIZE * tid];
 
   // Step 1: compute histogram
   for (int i = 0; i < RDX_HIST_SIZE; i++) {
@@ -713,10 +713,10 @@ void radix_sort_kernel(
     const int bin_3 = (key_3 >> (pass * 8)) & 0xFF;
     const int bin_4 = (key_4 >> (pass * 8)) & 0xFF;
 
-    const int pos_1 = local_histogram_ps[bin_1]++;
-    const int pos_2 = local_histogram_ps[bin_2]++;
-    const int pos_3 = local_histogram_ps[bin_3]++;
-    const int pos_4 = local_histogram_ps[bin_4]++;
+    const auto pos_1 = local_histogram_ps[bin_1]++;
+    const auto pos_2 = local_histogram_ps[bin_2]++;
+    const auto pos_3 = local_histogram_ps[bin_3]++;
+    const auto pos_4 = local_histogram_ps[bin_4]++;
 
     output_keys[pos_1] = key_1;
     output_values[pos_1] = input_values[i];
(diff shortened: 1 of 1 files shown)


DamianSzwichtenberg:
@q10 Could you please take a look at this PR? Some conversion error has slipped in the original `radix_sort` implementation. I tried to enable `-Wconversion` in the build system, but that has produced a ton of errors.

q10:
Hi @DamianSzwichtenberg thanks for opening this PR.  We currenrly have [another PR](https://github.com/pytorch/FBGEMM/pull/1710) in the pipeline, which fixes some build failures observed in this PR.  I will let you know to rebase this PR after we merge that other PR.  Thanks for your patience.

src/Utils.cc
```diff
@@ -652,17 +652,17 @@ void radix_sort_kernel(
...
+    int64_t* const histogram_ps,
     const int pass,
     const bool pass_with_sign_bit = false) {
   const auto tid = omp_get_thread_num();
   const auto nthreads = omp_get_num_threads();
   const auto elements_count_4 = elements_count / 4 * 4;
 
-  int* const local_histogram = &histogram[RDX_HIST_SIZE * tid];
-  int* const local_histogram_ps = &histogram_ps[RDX_HIST_SIZE * tid];
+  int64_t* const local_histogram = &histogram[RDX_HIST_SIZE * tid];
```
q10, line 664: Can we use `auto` here so that we won't have to worry about typing?

src/Utils.cc
```diff
@@ -756,14 +756,14 @@ std::pair<K*, V*> radix_sort_parallel(
 #ifdef _MSC_VER
   const size_t array_size = (size_t)RDX_HIST_SIZE * maxthreads;
   // fixes MSVC error C2131
-  int* const histogram = static_cast<int*>(
-      fbgemm::fbgemmAlignedAlloc(64, (array_size) * sizeof(int)));
-  int* const histogram_ps = static_cast<int*>(
-      fbgemm::fbgemmAlignedAlloc(64, (array_size + 1) * sizeof(int)));
+  int64_t* const histogram = static_cast<int64_t*>(
+      fbgemm::fbgemmAlignedAlloc(64, (array_size) * sizeof(int64_t)));
+  int64_t* const histogram_ps = static_cast<int64_t*>(
```
q10, line 761: same here

src/Utils.cc
```diff
@@ -756,14 +756,14 @@ std::pair<K*, V*> radix_sort_parallel(
...
+      fbgemm::fbgemmAlignedAlloc(64, (array_size + 1) * sizeof(int64_t)));
 
 #else
-  alignas(64) int histogram[RDX_HIST_SIZE * maxthreads];
-  alignas(64) int histogram_ps[RDX_HIST_SIZE * maxthreads + 1];
+  alignas(64) int64_t histogram[RDX_HIST_SIZE * maxthreads];
+  alignas(64) int64_t histogram_ps[RDX_HIST_SIZE * maxthreads + 1];
```
q10, line 766: same here

src/Utils.cc
```diff
@@ -713,10 +713,10 @@ void radix_sort_kernel(
...
     const int bin_4 = (key_4 >> (pass * 8)) & 0xFF;
 
-    const int pos_1 = local_histogram_ps[bin_1]++;
-    const int pos_2 = local_histogram_ps[bin_2]++;
-    const int pos_3 = local_histogram_ps[bin_3]++;
-    const int pos_4 = local_histogram_ps[bin_4]++;
+    const int64_t pos_1 = local_histogram_ps[bin_1]++;
+    const int64_t pos_2 = local_histogram_ps[bin_2]++;
+    const int64_t pos_3 = local_histogram_ps[bin_3]++;
+    const int64_t pos_4 = local_histogram_ps[bin_4]++;
```
q10, line 719: same here

q10:
@DamianSzwichtenberg You may now rebase this PR to top of tree

src/Utils.cc
DamianSzwichtenberg: Done.

src/Utils.cc
DamianSzwichtenberg: I cannot use auto here, as this is a declaration without initialization.

src/Utils.cc
DamianSzwichtenberg: Done. Though I would like to emphasize the difference between:
`const auto ptr = &val;  // equivalent to e.g. int64_t* const ptr = &val`
and
`const auto* ptr = &val; // equivalent to e.g. const int64_t* ptr = &val;`
I hope that code readers will not be confused by auto usage here.

DamianSzwichtenberg:
> @DamianSzwichtenberg You may now rebase this PR to top of tree

@q10 Changes rebased, suggestions applied. 😉

src/Utils.cc
```diff
@@ -756,14 +756,14 @@ std::pair<K*, V*> radix_sort_parallel(
 #ifdef _MSC_VER
   const size_t array_size = (size_t)RDX_HIST_SIZE * maxthreads;
   // fixes MSVC error C2131
-  int* const histogram = static_cast<int*>(
-      fbgemm::fbgemmAlignedAlloc(64, (array_size) * sizeof(int)));
-  int* const histogram_ps = static_cast<int*>(
-      fbgemm::fbgemmAlignedAlloc(64, (array_size + 1) * sizeof(int)));
+  const auto histogram = static_cast<int64_t*>(
```
q10, line 759: For pointer types, we should probably use `auto*` instead of plain `auto` for readability.

src/Utils.cc
```diff
@@ -652,17 +652,17 @@ void radix_sort_kernel(
...
     const int pass,
     const bool pass_with_sign_bit = false) {
   const auto tid = omp_get_thread_num();
   const auto nthreads = omp_get_num_threads();
   const auto elements_count_4 = elements_count / 4 * 4;
 
-  int* const local_histogram = &histogram[RDX_HIST_SIZE * tid];
-  int* const local_histogram_ps = &histogram_ps[RDX_HIST_SIZE * tid];
+  const auto local_histogram = &histogram[RDX_HIST_SIZE * tid];
+  const auto local_histogram_ps = &histogram_ps[RDX_HIST_SIZE * tid];
```
r-barnes, line 665:
```suggestion
  const auto* const local_histogram = &histogram[RDX_HIST_SIZE * tid];
  const auto* const local_histogram_ps = &histogram_ps[RDX_HIST_SIZE * tid];
```

src/Utils.cc
r-barnes, line 759:
```suggestion
  const auto* const histogram = static_cast<int64_t*>(
```

src/Utils.cc
```diff
@@ -756,14 +756,14 @@ std::pair<K*, V*> radix_sort_parallel(
 #ifdef _MSC_VER
   const size_t array_size = (size_t)RDX_HIST_SIZE * maxthreads;
   // fixes MSVC error C2131
-  int* const histogram = static_cast<int*>(
-      fbgemm::fbgemmAlignedAlloc(64, (array_size) * sizeof(int)));
-  int* const histogram_ps = static_cast<int*>(
-      fbgemm::fbgemmAlignedAlloc(64, (array_size + 1) * sizeof(int)));
+  const auto histogram = static_cast<int64_t*>(
+      fbgemm::fbgemmAlignedAlloc(64, (array_size) * sizeof(int64_t)));
+  const auto histogram_ps = static_cast<int64_t*>(
```
r-barnes, line 761:
```suggestion
  const auto* const histogram_ps = static_cast<int64_t*>(
```

r-barnes (changes requested):

src/Utils.cc
DamianSzwichtenberg: @q10 Yeah, good idea, I replaced `const auto` with `auto* const`.

src/Utils.cc
DamianSzwichtenberg: @r-barnes You probably meant here `auto* const` instead of `const auto* const`, which indeed is a better option than `const auto`. Code changed.

src/Utils.cc
DamianSzwichtenberg: As above.

src/Utils.cc
DamianSzwichtenberg: As above.

DamianSzwichtenberg:
@q10 PTAL.

PR #1721, opened 2023-04-25
Add CLI support (M,N,K) to GEMMsBenchmark

WilliamTambellini:
Add CLI support (M,N,K) to GEMMsBenchmark


bench/GEMMsBenchmark.cc
```diff
@@ -27,19 +27,23 @@
...
-    {16, 256, 512},
-    {128, 128, 128},
-    {256, 512, 256},
-    {1024, 1024, 1024},
+    {M?M:64, N?N:800, K?K:320},
+    {M?M:64, N?N:768, K?K:512},
+    {M?M:16, N?N:256, K?K:512},
+    {M?M:128, N?N:128, K?K:128},
+    {M?M:256, N?N:512, K?K:256},
+    {M?M:1024, N?N:1024, K?K:1024},
```
q10, line 44: We should parse `argc` and `argv` before calling `performance_test`, and just pass `M`, `N`, `K` to `performance_test` instead

WilliamTambellini: ok

WilliamTambellini: done

Issue #1947, opened 2023-08-17
Any plan to publish fbgemm_gpu that supports cuda 12.1?


henrylhtsang:
Right now while the wheel for cuda 12.1 is built and tested, it is not being published. Curious is there a plan to change that?

https://github.com/pytorch/FBGEMM/blob/5b0e0de2937f57af7993c097c27d8fb2fbee15c6/.github/workflows/fbgemm_gpu_cuda_nightly.yml#L133C43-L133C43

cc @osalpekar

shintaro-iwasaki:
Thanks for the question! Yes we test it for CUDA 12.1, but we don't publish the wheel, partly to avoid consuming too much PyPI storage (PyPI is not suitable for nightly). If we have a large space to upload it (with a good channel argument), we are very happy to publish all.

Currently we publish a wheel with the same CUDA version as what torch-nightly uses (#1954, cc @spcyppt).

osalpekar:
@shintaro-iwasaki We can store the wheel on `download.pytorch.org` which is backed by the same S3 index that's used by the rest of the PyTorch ecosystem. If we store all the nightlies there, you only need to push the release packages to PyPI, and we can deprecate use of the `fbgemm-*-nightly` packages entirely. This is what we are doing for torchrec, and should free up space to publish wheels for multiple CUDA versions.

spcyppt:
@osalpekar that sounds great, thank you.

colin2328:
Thanks @osalpekar !

shintaro-iwasaki:
@osalpekar Thank you! We are encountering a CUDA version mismatch issue, so definitely we want to store FBGEMM wheel files built with different CUDA versions soon!

Do you have any ETA in your mind? Can we help/do anything for this?
(cc: @spcyppt and @sryap)

osalpekar:
@shintaro-iwasaki No problem! @spcyppt reached out, and I sent over our [wiki](https://github.com/pytorch/test-infra/wiki/Using-Nova-Reusable-Build-Workflows) on how to migrate to our Nova pipelines, which should fix this issue as well as https://github.com/pytorch/FBGEMM/issues/1920. 

I'm happy to provide any help/support needed for the migration, maybe @spcyppt can give a better estimate for ETA but since we're not starting from scratch I don't think this should take too long?

shintaro-iwasaki:
Oh got it. Then I will ask her. Thank you very much for helping her on this work!

spcyppt:
[...]

PR #1958, opened 2023-08-19
Use Nova workflow to host all published wheel files at PyTorch site

spcyppt:
Summary:
To alleviate CUDA version mismatch issues, we publish fbgemm-gpu-nightly with different CUDA versions. This diff uses Nova workflow host the published wheel files at PyTorch site instead.

https://github.com/pytorch/FBGEMM/issues/1947

Differential Revision: D48487102


.github/workflows/build-wheels-linux.yml
```diff
@@ -0,0 +1,38 @@
...
+      with-cpu: enable
+  build:
+    needs: generate-matrix
+    name: pytorch/FBGEMM
+    uses: pytorch/test-infra/.github/workflows/build_wheels_linux.yml@main
+    with:
+      repository: pytorch/FBGEMM
+      ref: ""
+      pre-script: .github/script/nova_prescript.bash
+      post-script: .github/script/nova_prescript.bash
```
osalpekar, line 29: Should be `.github/scripts` instead of `.github/script`. This looks like the cause of the current breakage.

.github/workflows/build-wheels-linux.yml
```diff
@@ -0,0 +1,38 @@
...
+      smoke-test-script:
```
osalpekar, line 30: Let's explicitly make the value passed to this arg be ""

.github/scripts/nova_prescript.bash
```diff
@@ -0,0 +1,25 @@
...
+#Display GPU Info
+$PRELUDE; print_gpu_info
+#Install C/C++ Compilers
+$PRELUDE; install_cxx_compiler $BUILD_ENV
+#Install Build Tools
+$PRELUDE; install_build_tools $BUILD_ENV
+#Install CUDA
+$PRELUDE; install_cuda $BUILD_ENV "$CU_VERSION"
+# Install PyTorch Nightly
+$PRELUDE; install_pytorch_pip $BUILD_ENV nightly cuda "$CU_VERSION"
```
osalpekar, line 21: CUDA and the right pytorch version should already be installed for you under the hood by our pipeline, there is no need to do so in your pre-script.

osalpekar (commented):
Generally looks good! Left a few nits, and might have to iterate a bit to get the builds passing in CI

.github/workflows/build-wheels-linux.yml
spcyppt: Thank you for catching that! I have fixed it.

.github/workflows/build-wheels-linux.yml
spcyppt: Thank you, done.

.github/scripts/nova_prescript.bash
spcyppt: Thank you so much, I've removed them.

atalman:
HI @spcyppt please fix this failure: https://github.com/pytorch/FBGEMM/actions/runs/5908914110/job/16028907266?pr=1958

spcyppt:
Hi @atalman, sorry for pinging you again here. The build fails as it can't locate the file `setup.py`, which is in fbgemm_gpu subfolder. I did `cd` to the folder in my prescript but I guess the working directory is reset at the build clean step. Please let me know how to fix this. Thank you. https://github.com/pytorch/FBGEMM/actions/runs/5931552597/job/16083645238#step:13:54

.github/scripts/nova_postscript.bash
```diff
@@ -0,0 +1,24 @@
...
+echo "Current working directory: $(pwd)"
+cd "$FBGEMM_REPO" || echo "Failed to cd to $FBGEMM_REPO"
+PRELUDE=.github/scripts/setup_env.bash
+BUILD_ENV_NAME=base
+
+# Install FBGEMM_GPU Nightly
+echo "Current working directory: $(pwd)"
+# shellcheck disable=SC1091
+# shellcheck source=.github/scripts/setup_env.bash
+. $PRELUDE; install_fbgemm_gpu_package "$BUILD_ENV_NAME" fbgemm_gpu/dist/*.whl
```
osalpekar, line 17: Looks like the job is now failing in the post-script phase. Is this script looking for the wheel in the incorrect directory?

spcyppt: The directory is correct, since I already cd "$FBGEMM_REPO". The wheel should be in fbgemm_gpu/dist. I think the problem is the building wheel step, so I'm fixing that.

PR #2001, opened 2023-09-06
for future rocm versions

liligwu:
For upcoming ROCm versions, the header file location will change.


.github/scripts/fbgemm_gpu_build.bash
```diff
@@ -444,6 +444,24 @@ build_fbgemm_gpu_develop () {
...
+/usr/local/sbin:\
+/usr/local/bin:\
+/usr/sbin:\
+/usr/bin:\
+/sbin:\
+/bin"
+    echo $new_path
+    print_exec conda env config vars set ${env_prefix} PATH="$new_path"
+    (exec_with_retries 3 conda run ${env_prefix} echo "===============$PATH") || return 1
+  fi
```
q10, line 464: You may want to define this configuration inside `__configure_fbgemm_gpu_build_rocm` instead

PR #2024, opened 2023-09-15
Add aarch64 builds on Nova

spcyppt:
Summary: Add arm builds through Nova to be hosted on pytorch.org

Differential Revision: D49333436

### DIFF
--- a/.github/workflows/build_wheels_aarch64_linux.yml
+++ b/.github/workflows/build_wheels_aarch64_linux.yml
@@ -1,0 +1,60 @@
+name: Build Aarch64 Linux Wheels
+
+on:
+    pull_request:
+    push:
+      branches:
+        - nightly
+        - main
+        # Release candidate branch look like: v1.11.0-release
+        - v[0-9]+.[0-9]+.[0-9]+-release+
+      tags:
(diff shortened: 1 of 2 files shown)


.github/workflows/build_wheels_aarch64_linux.yml
```diff
@@ -0,0 +1,60 @@
...
+  build:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - repository: pytorch/FBGEMM
+            smoke-test-script: ""
+            pre-script: ../.github/scripts/nova_prescript.bash
+            post-script: ../.github/scripts/nova_postscript.bash\
```
osalpekar, line 39: nit: remove trailing `\`. also don't need the `../` since it starts looking in the project root directory

osalpekar (approved):
workflow LGTM

.github/workflows/build_wheels_aarch64_linux.yml
spcyppt: I force `cd fbgemm_gpu` in `env-var-script` if it's in project root directory. Will the script gets read before or after the `env-var-script` is executed?

PR #2059, opened 2023-10-05
Fix inconsistent dll linkage warning

cyyever:


cyyever:
@q10 Now it should work.

PR #2067, opened 2023-10-06
Fix MSVC 'value of intrinsic immediate argument out of range'

cyyever:
MSVC reports
```
 warning C4556: value of intrinsic immediate argument '8' is out of range '0 - 7'
```
Fix it according to the suggestion from https://developercommunity.visualstudio.com/t/-mm-cvtps-ph-doesnt-accept-mm-fround-no-exc/1343857


q10:
Looks like the changes are causing test failures

cyyever:
Yes, I would take some time to investigate

PR #2086, opened 2023-10-23
[ROCm] updates for ROCm 6.0 support

jeffdaily:
ROCm 6.0 introduces backwards-incompatible changes such as removing the long-deprecated use of `__HIP_PLATFORM_HCC__`. It is better to use the USE_ROCM macro which is already defined and indicates a ROCm build. This PR also defines `__HIP_PLATFORM_AMD__` which is the new symbol name. This symbol is still required for compiling with HIP headers but when not using hip-clang.


jeffdaily:
@sryap what are the Facebook Internal Linter failures?

PR #2100, opened 2023-10-27
Delete copy and move operations of some Matrix classes

cyyever:
For those matrices managing their buffers, the copy and move operations should be deleted to avoid default versions generated by the compiler. Another solution is to manage the buffer ownership in user defined versions, but deleting them is simpler given that no memory corruption has been found since it can be inferred that they are not used.

Another solution is use std::share_ptr to manager buffers. However, it will incur some atomic operations on the reference counter, which is not desirable .


q10:
@cyyever can I trouble you rebase the branch to top of tree `main`?

cyyever:
@q10 Done

PR #2106, opened 2023-11-01
simplify Hip.cmake and prepare for the future ROCm versions

liligwu:


shintaro-iwasaki:
Thanks! Looks good to me. Let me merge it.

PR #2153, opened 2023-11-21
[fbgemm_gpu] Disable CXX11 ABI by default

q10:
- Disable compilation to CXX11 ABI by default because the built version of PyTorch library that is available on PIP is compiled without CXX11 ABI


q10:
The issue this PR was intended to fix turns out to be something else, so will close the PR

PR #2235, opened 2023-12-27
Fix fbgemm TBE building issues

jianyuh:
Summary: Unblock of fbgemm TBE (inference, training) usages on AMD GPUs.

Differential Revision: D52425243


houseroad (approved):
Please ensure the CI is green. Looks like some failure are related?

PR #2269, opened 2024-01-16
Make auto select for embedding_ops and fix duplicate cu files

houseroad:
Summary:
Both "embedding_ops" and "index_select_ops" include "gen_batch_index_select_dim0_backward_kernel_warp.cu", which caused the duplicate symbol issue.

So remove the "gen_embedding_backward_split_grad.cu" from the source of "index_select_ops", and include "embedding_ops" as its deps.

Reviewed By: jiaqizhai

Differential Revision: D52790445


houseroad:
The failed tests look unrelated.

PR #2353, opened 2024-02-26
Use index-url when installing from pip

atalman:
See install instructions in: https://pytorch.org/get-started/locally/


flaviotruzzi (approved):
LGTM

PR #2436, opened 2024-03-18
FP32 autovec implementation

crystalrchen:


tall fbgemm cpu
excelle08, line 1: What's the purpose of this file? We should remove this if it's not a part of the code.

PR #2489, opened 2024-04-09
Use release branch of Nova workflow

spcyppt:
Differential Revision: D55947486

### DIFF
--- a/.github/workflows/build_wheels_linux_aarch64.yml
+++ b/.github/workflows/build_wheels_linux_aarch64.yml
@@ -24,7 +24,7 @@ permissions:
 
 jobs:
   generate-matrix:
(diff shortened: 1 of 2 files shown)


huydhn (approved):

huydhn:
Wait a sec, I thought this is going to fbgemm release branch.  For the main branch, I think it's better to use the test-infra @main instead. And yes, we will add the option to fallback to `bdist_wheel`.  What I mean is that fbgemm release branch can use test-infra @release/2.3.  This is what we call a release-only change on PyTorch.

spcyppt:
@huydhn It's also currently blocking our nightly releases, so I'm changing this in main to unblock nightly release. Once the fallback is added and merged, I will change back to main.

PR #2347, opened 2024-02-22
- Add codegen for embedding backward meta functions

q10:
Summary:
Adding embedding backward meta codegen functions.

Moved memory alignment that was outside of the cuda kernel into the custom operator, since we couldn't write a symbolic version for memory alignment checks on the pointers.

Tests are changed to allow compilation only on adagrad. Other tests are ran to ensure they continue to work properly.

There are missing fixes to allow compilation for unweighted kernels and CPU, which are excluded from the tests.

Differential Revision: D53674518

PR #2526, opened 2024-04-22
Avoid cloning gradient tensor in embedding backward pass

jhadidjojo:
Summary: I found memory spike during embedding kernel backward `split_embedding_backward_codegen_rowwise_adagrad_unweghted_exact_cuda`, which was traced into the below code making a clone of the gradient tensor. This logic didn't seem to be there in the original code: https://github.com/pytorch/FBGEMM/pull/2347/files#diff-944ab49dcbcf54826cc3e1eab5e3c0c787b5a195f602c2d3052adae14c506d78.

Differential Revision: D56420646

### DIFF
--- a/fbgemm_gpu/include/fbgemm_gpu/sparse_ops_utils.h
+++ b/fbgemm_gpu/include/fbgemm_gpu/sparse_ops_utils.h
@@ -466,7 +466,7 @@ struct StackArray {
 
 inline at::Tensor aligned_grad_output_tensor_for_cuda_backwards(
     const at::Tensor& grad_output) {
-  auto aligned_grad_output = grad_output.clone();
+  auto aligned_grad_output = grad_output;
   // FIXME: to support aligned memory access in Vec4T load/store function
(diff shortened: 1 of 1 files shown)


flaviotruzzi:
This will break the signature for const, will it not?

jhadidjojo:
> This will break the signature for const, will it not?

No, I don't think so. You're right that `grad_output` is a const reference, and if you try to assign that to a new value it will error. Here I'm creating a new (non-const) reference called `aligned_grad_output`. You can assign it to the `grad_output` by reference without copy.

flaviotruzzi:
> > This will break the signature for const, will it not?
> 
> No, I don't think so. You're right that `grad_output` is a const reference, and if you try to assign that to a new value it will error (that's probably the error you saw?). Here I'm creating a new (non-const) reference called `aligned_grad_output`. You can assign it to the `grad_output` by reference without copy.

Cool, I am only on my phone, and still on leave, can you check with someone else to approve it?

PR #2582, opened 2024-05-10
FP8 AMD Kernels

jwfromm:
Summary:
AMD support for the `f8f8bf16` fbgemm kernel. This implementation uses the Composable Kernel library to construct a performant FP8 Gemm on AMD. It runs up to 10X faster than `scaled_mm`.

Benchmarking data is available [here](https://docs.google.com/spreadsheets/d/1hK62EJIt9mZQJdZBxf5rGZ3u0hjMng7AkpwKnIjyn3k/edit#gid=500854639). Note that I primarily focused on larger workloads. On small workloads, there doesnt seem to be much benefit to using fp8 over bf16.

Differential Revision: D56963018

### DIFF
--- a/fbgemm_gpu/CMakeLists.txt
+++ b/fbgemm_gpu/CMakeLists.txt
@@ -98,7 +98,7 @@ if(NOT FBGEMM_CPU_ONLY)
   add_subdirectory(experimental/example)
 
   if(NOT USE_ROCM)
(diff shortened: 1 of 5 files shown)


jeffdaily:
How do you build and benchmark this if the experimental/gen_ai subdir is excluded for ROCm builds?

PR #2586, opened 2024-05-13
FP32 Autovec Final Optimization

crystalrchen:
1. Verified that FP32 implementation passes unit tests
2. Created FP32 RowWiseSparse function (EmbeddingSpMDMRowWiseSparse_autovec)
3. Optimized FP32 Autovec Function (EmbeddingSpMDM_autovec):
    * #pragma omp tile size (tile_size = 4)
    * cache prefetching (max_initial_prefetch_row = 8)
    * #pragma omp simd
4. Implemented switching logic for usage of the autovec environment variable for both EmbeddingSpMDM_ref/autovec and EmbeddingSpMDMRowWiseSparse_ref/autovec


src/EmbeddingSpMDMAutovec.cc
```diff
@@ -435,6 +435,483 @@ bool EmbeddingSpMDMNBit_autovec(
...
+    bool is_weight_positional /*=false*/,
+    bool use_offsets /*=true*/,
+    int64_t output_stride /*=-1*/,
+    int64_t input_stride /*=-1*/,
+    bool scale_bias_last /*=true*/,
+    bool no_bag /*=false*/,
+    bool is_bf16_out /*=false*/,
+    bool is_bf16_in /*=false*/) {
+  const bool isWeight8bit = std::is_same<InType, uint8_t>::value;
+  const bool isOutput8bit = std::is_same<OutType, uint8_t>::value;
```
jspark1105, line 463: nit: constexpr?

src/EmbeddingSpMDMAutovec.cc
```diff
@@ -435,6 +435,483 @@ bool EmbeddingSpMDMNBit_autovec(
...
+  if (output_stride == -1) {
+    output_stride = block_size;
+  }
+  if constexpr (isOutput8bit) {
+    assert(input_stride == output_stride);
+  }
+  vector<float> buf(block_size);
+
+  if (isWeight8bit) {
```
jspark1105, line 472: constexpr

src/EmbeddingSpMDMAutovec.cc
```diff
@@ -435,6 +435,483 @@ bool EmbeddingSpMDMNBit_autovec(
...
+        int64_t idx = indices[m];
+
+        if (idx < 0 || idx >= data_size) {
+          return false;
+        }
+        if constexpr (isOutput8bit) {
+          const InType* input_row_ptr = input + input_stride * idx;
+          memcpy(out, input_row_ptr, sizeof(InType) * input_stride);
+        } else {
+          memset(buf.data(), 0, sizeof(float) * block_size);
```
jspark1105, line 495: Is initializing to zero and using fma helps vectorization vs. leave with garbage values and overwrite with multiplication results?

excelle08: Yeah initializing to zero is not really necessary for sequential embeddings because there is no aggregation from multiple embedding rows. It is required in pooled embeddings for correctness, though.

PR #2693, opened 2024-06-06
[comm][ROCm] move memory copy into one_shot_all_reduce

wenkaidu:
Avoid latency of launching hipMemcpyAsync. Could see 3-4us reduction in benchmarking. Also see improvements in end to end testing.

### DIFF
--- a/fbgemm_gpu/experimental/gen_ai/src/comm/car.cu
+++ b/fbgemm_gpu/experimental/gen_ai/src/comm/car.cu
@@ -139,9 +139,20 @@ __global__ void one_shot_all_reduce(
     int32_t flag,
     std::array<int32_t*, 8> barriers,
     std::array<at::BFloat16*, 8> inputs,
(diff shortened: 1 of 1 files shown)


xw285cornell:
Thanks @wenkaidu ! Just wondering, what's this diff's impact on CUDA/H100? will that be generally beneficial?

wenkaidu:
> Thanks @wenkaidu ! Just wondering, what's this diff's impact on CUDA/H100? will that be generally beneficial?

I heard cudaMemcpy is more performant than copy from kernel. But I am not sure which method is faster with small data sizes. I don't have access to H100 HW and SW setup. If someone have setup, please run a quick check.

xw285cornell (approved):
nice thanks for the contribution!

xw285cornell:
@wenkaidu sorry I have to create another PR #2770 to fix some formatting issue (and I cannot re-export this PR). Thanks again for your contribution!

PR #2716, opened 2024-06-11
Prefer torch.compiler.is_compiling to torch._dynamo.is_compiling

fbgheith:
Summary:
`torch._dynamo.is_compiling` is deprecated and it just forwards to `torch.compiler.is_compiling`: https://www.internalfb.com/code/fbsource/[99ea324119d7]/fbcode/caffe2/torch/_dynamo/external_utils.py?lines=28

What's the urgency? two `pytorch` PR's had to be reverted in the past week because using `torch._dynamic.is_comiling` indirectly internal (to Meta) failures.

The failure scenario involves using `jit.script.compile` which is not decorator friendly and deprecating  `torch._dynamo.is_compiling` involves decorating the function with `dprecated`.

There are viable alternatives that involve more effort and risk:

* "fix" `jit.script.compile`
* don't use `jit.script.compile`
* don't decorate `torch._dynamic.is_compiling`

Review question:
* should this be done on the OSS side?

Differential Revision: D58415423

### DIFF
--- a/fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops_training.py
+++ b/fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops_training.py
@@ -60,7 +60,16 @@ except Exception:
 
 
 try:
-    from torch._dynamo import is_compiling as is_torchdynamo_compiling
+    try:
+        from torch.compiler import is_compiling
+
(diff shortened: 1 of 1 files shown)


clee2000:
This should help with https://github.com/pytorch/pytorch/pull/127690

PR #2822, opened 2024-07-11
Remove redundant torch.abs in sim check

acoskunses-AMD:
Summary:
The absolute value is redundant because
the next step involves squaring the differences,
which makes all values non-negative.


acoskunses-AMD:
@jianyuh @jwfromm

PR #2895, opened 2024-07-25
Replace runners prefix amz2023.

jeanschmidt:
testing new runners


q10:
Hi @jeanschmidt It looks iike some of the runners are not available.  Since the PR was made 3 months ago Is it still relevant, or should we close it out?

PR #2610, opened 2024-05-20
[ROCm] enable experimental gen_ai build

jeffdaily:
Follow up to #2582.

CC @jwfromm

PR #2928, opened 2024-08-01
[fbgemm_gpu] HIP support for GenAI ops

q10:


mht-sharma:
@q10 , I believe we also need to integrate or install the Composable Kernel (CK) for the GenAI ops. Were you able to build with CK? If so, could you please share the steps you followed? I’m running into some issues and would greatly appreciate any guidance you can provide.

cc @jeffdaily, I’ve seen similar [PR 2610](https://github.com/pytorch/FBGEMM/pull/2610) from you and thought you might have some insights as well.

q10:
> @q10 , I believe we also need to integrate or install the Composable Kernel (CK) for the GenAI ops. Were you able to build with CK? If so, could you please share the steps you followed? I’m running into some issues and would greatly appreciate any guidance you can provide.
> 
> cc @jeffdaily, I’ve seen similar [PR 2610](https://github.com/pytorch/FBGEMM/pull/2610) from you and thought you might have some insights as well.

Ah yes, thanks for the pointer on CK.  This work has stalled a bit due to other priorities, but ROCm support for GenAi ops is a work in progress.

PR #3047, opened 2024-08-28
TorchRec MX4 padding support

qchip:
Summary:
X-link: https://github.com/pytorch/torchrec/pull/2344

rowwise padding is needed to fix NE of Ads ranking models for AlltoAll communication with MX4.

Differential Revision: D61928939


qchip:
split and landed in https://github.com/pytorch/FBGEMM/pull/3055

PR #3127, opened 2024-09-12
Fp8 hipstream fix

acoskunses-AMD:
Pick hip stream where cuda is not defined

### DIFF
--- a/fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/fp8_blockwise_gemm.hip
+++ b/fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/fp8_blockwise_gemm.hip
@@ -15,8 +15,6 @@
 #include <c10/hip/HIPStream.h>
 #include <torch/torch.h>
 
-#if defined(USE_ROCM)
-
 #include "ck/ck.hpp"
 #include "ck/tensor_operation/gpu/device/gemm_specialization.hpp"
 #include "ck/tensor_operation/gpu/device/tensor_layout.hpp"
@@ -272,5 +270,3 @@ at::Tensor f8f8bf16_blockwise(
 }
 
 } // namespace fbgemm_gpu
-
-#endif // defined(USE_ROCM)
--- a/fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/kernels/fp8_rowwise_common.h
+++ b/fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/kernels/fp8_rowwise_common.h
@@ -15,8 +15,6 @@
 #include <c10/cuda/CUDAStream.h>
 #include <torch/torch.h>
 
-#if defined(USE_ROCM)
-
 #include "ck/ck.hpp"
 #include "ck/tensor_operation/gpu/device/gemm_specialization.hpp"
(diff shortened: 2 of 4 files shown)


acoskunses-AMD:
@jianyuh @jwfromm

xw285cornell:
We use hipify script for this, is the change here needed?

jianyuh:
Rebase to resolve conflicts: also as Xiaodong mentioned, internally we don't need this (automatically hipify). Is this needed on OSS workflow? I guess hipify torch is not enabled?

fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/kernels/fp8_rowwise_common.h
```diff
@@ -12,10 +12,13 @@
 #include <numeric>
 
 #include <ATen/ATen.h>
+#if !defined(USE_ROCM)
```
xw285cornell, line 15: hipify script should hipify .h right?

fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/fp8_blockwise_gemm.hip
```diff
@@ -12,11 +12,14 @@
 #include <numeric>
 
 #include <ATen/ATen.h>
-#include <c10/hip/HIPStream.h>
+#if !defined(USE_ROCM)
```
xw285cornell, line 15: is this block even needed? If USE_ROCM is not defined, it'll basically be an empty file. So we can just remove this block here.

fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/kernels/fp8_rowwise_common.h
xw285cornell: in any case, I think maybe we can just remove this block just to be clean.

fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/fp8_blockwise_gemm.hip
xw285cornell: Let's just remove this include and move it down to the USE_ROCM. No need to include cuda header here.

acoskunses-AMD: @xw285cornell  @jianyuh 

changed as required. Please review.

xw285cornell (approved):
I guess we should never use nvcc to build those files so we should be ok?

fbgemm_gpu/experimental/gen_ai/src/quantize/ck_extensions/kernels/fp8_rowwise_common.h
```diff
@@ -201,4 +199,3 @@ at::Tensor f8f8bf16_rowwise_impl(
   return Y;
 }
```
xw285cornell, line 201: @acoskunses-AMD can you remove the blank lines at the end of each file? Otherwise we cannot land it internally due to some lint check

PR #3159, opened 2024-09-21
[fbgemm_gpu] Add workflow for running only on non-PyTorch infrastructure

q10:
- Add workflow for running only on non-PyTorch infrastructure


basilwong:
Added comments to internal version

PR #3266, opened 2024-10-22
[ROCm] debug v2 kernel for ROCm

liligwu:
Fix the v2 kernel forward_test for ROCm. One more unit test failure (test_forward_gpu_uvm_cache_fp16) is under investigation.

### DIFF
--- a/fbgemm_gpu/codegen/training/forward/embedding_forward_split_kernel_v2_template.cu
+++ b/fbgemm_gpu/codegen/training/forward/embedding_forward_split_kernel_v2_template.cu
@@ -794,7 +794,7 @@ __global__ void split_embedding_codegen_forward_{{ wdesc }}_v2_kernel(
       return;
     }
 
-    uint32_t load_D;
+    uint32_t load_D = 0;
     uint32_t D_start;
     uint32_t total_load_D;
 
@@ -814,7 +814,7 @@ __global__ void split_embedding_codegen_forward_{{ wdesc }}_v2_kernel(
     const uint32_t load_d = (table_warp_id % num_warps_per_row) * kWarpSize;
     // Compute sample ID
     const uint32_t b = table_warp_id / num_warps_per_row * (is_small_L ? NUM_OFFSETS_PER_WARP : 1);
-    uint32_t L;
+    uint32_t L = 0;
     uint32_t row_start;
     bool use_lxu_cache = USE_LXU_CACHE;
 
@@ -896,7 +896,7 @@ __global__ void split_embedding_codegen_forward_{{ wdesc }}_v2_kernel(
     }
 
     if (is_small_L) {
-      INVOKE_PROCESS_ALL_INDICES(small_Ls, 32, 0xf)
+      INVOKE_PROCESS_ALL_INDICES(small_Ls, kWarpSize, 0xf)
       return;
     }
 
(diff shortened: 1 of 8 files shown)


xw285cornell (commented):
Nice! Can you publish this?

jspark1105:
Thanks! Is it correct we can deprecate ``is_rocm`` parameter in ``has_experimental_support``? Don't mean we should do this in this PR and just wanted to confirm so we can work on a follow-up PR.

avbokovoy:
@liligwu 
> Fix the v2 kernel forward_test for ROCm. One more unit test failure (test_forward_gpu_uvm_cache_fp16) is under investigation.

[c674210](https://github.com/pytorch/FBGEMM/pull/3266/commits/c67421034db930f20fc2f61ca146b4257f4dd6e1) should fix **uvm_cache_fp16** test

@jspark1105 
> Thanks! Is it correct we can deprecate `is_rocm` parameter in `has_experimental_support`? Don't mean we should do this in this PR and just wanted to confirm so we can work on a follow-up PR.

I don't see any significant reasons to keep this parameter, unless some new experimental features will be added without ROCm support. If that's the case, you may just revert [94b3517](https://github.com/pytorch/FBGEMM/pull/3266/commits/94b35179ebb7109753b7f66338b6586a52a88011)

fbgemm_gpu/codegen/training/forward/embedding_forward_split_kernel_v2_template.cu
```diff
@@ -738,6 +738,12 @@ __global__ void split_embedding_codegen_forward_{{ wdesc }}_v2_kernel(
     using emb_vec_t = vec4_type<emb_t>;
     using cache_vec_t = vec4_type<cache_t>;
     using output_vec_t = vec4_type<output_t>;
+#if defined(USE_ROCM)
+    using look_up_bits_t = uint64_t;
```
xw285cornell, line 742: why it needs to be uint64_t for rocm?

fbgemm_gpu/include/fbgemm_gpu/utils/cuda_prelude.cuh
```diff
@@ -73,7 +77,7 @@ DEVICE_INLINE T shfl_xor(
     const T val,
     int laneMask,
     int width = kWarpSize,
-    unsigned shfl_sync_mask = kFullWarpMask) {
+    unsigned shfl_sync_mask = static_cast<unsigned>(kFullWarpMask)) {
```
xw285cornell, line 80: Do we need explicit cast here? And isn't unsigned == unsigned int, why did we define it as uint64_t above?

fbgemm_gpu/codegen/training/forward/embedding_forward_split_kernel_v2_template.cu
avbokovoy: This is done to account for ROCm 64 wave size (particularly, to get the right result from [bullot_sync](https://github.com/pytorch/FBGEMM/blob/89de1da1583303db0da9fbf7d9a3133ac7159ae1/fbgemm_gpu/include/fbgemm_gpu/utils/cuda_prelude.cuh#L111) function).

fbgemm_gpu/include/fbgemm_gpu/utils/cuda_prelude.cuh
avbokovoy: This is done to avoid massive "implicit-cast" warning spam if built with USE_ROCM. **shfl_sync_mask** is not used in this case anyway (why - is another question). Regarding uint64_t - see the comment [above](https://github.com/pytorch/FBGEMM/pull/3266#discussion_r1818558455)

xw285cornell (approved):

PR #3309, opened 2024-11-01
Add manual loop unroll for rocm devices in fwd pass

avbokovoy:
Added another instance of FWD pass for ROCm devices. Manually split load and accumulate/store macros and manually unrolled outer loops. Added run-time guard check for ROCm devices. Current limitation is L % 4 == 0.

Some of forward unit tests fail (probably due to guard check), need to investigate that. Until UT fixed, keeping this PR as draft.

Performance measurements will be shared via another channel.

cc: @liligwu @amathews-amd


avbokovoy:
Found a way to add L % 4 != 0 support and significantly reduce diff needed to upstream optimization.

PR #3421, opened 2024-11-27
Add sve implementation for float matrix transpose

Nicoshev:
Summary: Adding sve-based function for transposing float matrixes

Differential Revision: D66528598


snadampal:
Hi @Nicoshev , are these kernels under `USE_FBGEMM` flag in PyTorch?
last time when I tried to build PyTorch with FBGEMM backend on aarch64 I ran into multiple issues due to `avx2` and `avx512` hardcoded logic under this flag.  is it supported now?

PR #3424, opened 2024-11-28
Add NEON and SVE implementations for Float16 conversions

annop-w:


psaab:
```
diff --git a/fbcode/deeplearning/fbgemm/TARGETS b/fbcode/deeplearning/fbgemm/TARGETS
--- a/fbcode/deeplearning/fbgemm/TARGETS
+++ b/fbcode/deeplearning/fbgemm/TARGETS
@@ -184,6 +184,7 @@
         "-DFBCODE_CAFFE2",
     ],
     exported_deps = [
+        ":fbgemm_base",
         "fbsource//third-party/asmjit:asmjit",
         "fbsource//third-party/cpuinfo:cpuinfo",
         "//common/aarch64:intrinsics_dep",
diff --git a/fbcode/deeplearning/fbgemm/defs.bzl b/fbcode/deeplearning/fbgemm/defs.bzl
--- a/fbcode/deeplearning/fbgemm/defs.bzl
+++ b/fbcode/deeplearning/fbgemm/defs.bzl
@@ -143,7 +143,11 @@
     return asm_srcs if not msvc else intrinsics_srcs
 
 def get_fbgemm_inline_sve_srcs(msvc = False, buck = False):
-    intrinsics_srcs = ["src/FbgemmFP16UKernelsSve128.cc"]
+    intrinsics_srcs = [
+        "src/FbgemmFP16UKernelsSve128.cc",
+        "src/FbgemmFloat16ConvertNeon.cc",
+        "src/FbgemmFloat16ConvertSve.cc"
+    ]
 
     #FP16 kernels contain inline assembly and inline assembly syntax for MSVC is different.
     asm_srcs = ["src/FbgemmFP16UKernelsSve128.cc"]
```

embg:
Hey, I was not aware of this PR until about an hour ago. It seems that my change https://github.com/pytorch/FBGEMM/pull/3626 covers some of the same ground. Sorry about that.

I'm going to go ahead an land my change, because we need it urgently for one of our big workloads, and it would take a bit of time to import this PR and fix it up for landing internally.

However I think that this PR is great work, and adds a lot of value beyond my change, namely:
* support for NEON
* support for SVE (I only did SVE2)
* support for Float16ToFloat (I only did the other direction, because upcast isn't hot in the workloads we've profiled so far)

I'm happy to work with you to land this, if you are interested, after https://github.com/pytorch/FBGEMM/pull/3626 lands.

embg:
Btw, I notice that in your PR you don't do anything related to FP rounding mode. I actually tried to do some pragmas in my PR, but I'm 99% sure the compiler ignores them (at least based on some Godbolt experiments).

Do you have any advice in general on how to handle the FP environment stuff when writing SVE2 code?

For AVX, the rounding mode is passed as an immediate value, so it's not an issue. But I got scared when I read in the SVE2 docs that the rounding mode is based on the FP environment.

e.g. https://godbolt.org/z/hn1zqf1dT

annop-w:
> I'm happy to work with you to land this, if you are interested, after #3626 lands.
Absolutely. Please let me know how you'd like to proceed.

annop-w:
> Btw, I notice that in your PR you don't do anything related to FP rounding mode.
The rounding behavior for the FCVT instruction follows the current rounding mode set in the FPCR register:

https://developer.arm.com/documentation/ddi0502/g/programmers-model/aarch64-register-descriptions/floating-point-control-register

One can use fegetround and fesetround to query/control the rounding behavior.

https://en.cppreference.com/w/c/numeric/fenv/feround

annop-w:
I also would like to mention that I have similar implementations for BFloat16 <-> Float conversions locally. I can open another PR when this Float16 version goes in. 

May I also bring your attention to another PR for transpose kernel ? #3510

embg:
> One can use fegetround and fesetround to query/control the rounding behavior.

So here's the part that's making things tricky. On cppreference, [they say say you can't use fesetround without certain pragmas](https://en.cppreference.com/w/cpp/numeric/fenv):

> The floating-point environment access and modification is only meaningful when [#pragma STDC FENV_ACCESS](https://en.cppreference.com/w/cpp/preprocessor/impl) is supported and is set to ON.

The problem is, clang does *not* like those pragmas. I had to remove one from my previous PR: https://github.com/pytorch/FBGEMM/pull/3629

Maybe the answer is -- just use `fesetround()` without any pragmas? I need to dig a bit deeper on this. If you have any insights, please share.

In terms of proceeding with this merge, I'd like to figure out the floating-point environment situation, fix up my previous PR according to what we figure out, and then we can apply the same fixes to your code.

embg:
In the meantime, could you benchmark your SVE FloatToFloat16 against my SVE2 implementation (merged)? I used `ld2w` and you used `ld1w`, but you did loop unrolling, so I'm not sure which one is faster.

annop-w:
> Maybe the answer is -- just use fesetround() without any pragmas? I need to dig a bit deeper on this. If you have any insights, please share.

I have tried with a simple program with GNU compiler without the pragma and it works OK. I could try with clang to confirm.

> In the meantime, could you benchmark your SVE FloatToFloat16 against my SVE2 implementation (merged)? I used ld2w and you used ld1w, but you did loop unrolling, so I'm not sure which one is faster.

Yes, I will let you know the results.

embg:
> May I also bring your attention to another PR for transpose kernel ? https://github.com/pytorch/FBGEMM/pull/3510

I pinged someone internally on this

annop-w:
> Btw, I am benchmarking on Grace, if your hardware is different please flag that

I will do the benchmark on Grace-Grace system then.

annop-w:
Here is the benchmark result:
```
   M           elements_per_ns
         ref   SVE2     SVE    NEON
   1  0.0257  0.025  0.0256  0.0233
  10   0.196  0.225   0.227    0.22
  32   0.322  0.511   0.538   0.534
  40   0.342  0.584    0.61   0.618
 129   0.381  0.921    0.95   0.967
 256   0.403   1.04    1.09    1.08
1024   0.415    1.2    1.24    1.23
8000   0.419   1.26    1.28    1.27
```

PR #3506, opened 2024-12-15
Back out "Manual loop unroll for rocm inference"

brad-mengchi:
Summary:
Original commit changeset: 66ba86adbd5e

Original Phabricator Diff: D66563556

Differential Revision: D67246937


q10:
Context for the AMD folks:

There was an internal perf benchmark that regressed and seemed to have pointed to PR 3506, presumably because the code was not tuned for the nobag TBE case.  However, after some discussions, it was concluded that the regression is within noise, and that this revert PR will not be landed until more detailed tests confirm this hypothesis.

PR #3528, opened 2024-12-23
Revert "Attempt to fix FBGEMM CPU build (#3499)"

huydhn:
This reverts commit 5c16f4b22afb32e0bf4817f5210f0614bb0fb17a.  This is not needed anymore after https://github.com/pytorch/pytorch/pull/143423.  I think this will also fix the issue with building torchrec CPU https://github.com/pytorch/FBGEMM/actions/runs/12470608879/job/34806045264?pr=3528#step:18:219⁩

### Testing

https://github.com/pytorch/FBGEMM/actions/runs/12470608879

### DIFF
--- a/.github/scripts/nova_dir.bash
+++ b/.github/scripts/nova_dir.bash
@@ -13,9 +13,6 @@ if [[ "$working_dir" == "$FBGEMM_REPO" ]]; then cd fbgemm_gpu || echo "Failed to
 ## Build clean/wheel will be done in pre-script. Set flag such that setup.py will skip these steps in Nova workflow
 export BUILD_FROM_NOVA=1
 
-## TODO (huydhn): Remove this when Nova is moving to manywheel 2.28
-export SET_GLIBCXX_PRELOAD=1
-
 ## Overwrite existing ENV VAR in Nova
 if [[ "$CONDA_ENV" != "" ]]; then export CONDA_RUN="conda run --no-capture-output -p ${CONDA_ENV}" && echo "$CONDA_RUN"; fi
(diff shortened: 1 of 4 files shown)


huydhn:
Hmm, CPU build fails when this is reverted, so maybe we need this after all

atalman (approved):
lgtm

q10 (approved):

huydhn:
I think I get this working for all linux builds now, but the fix is more complicated than I thought.  We need to:

1. Use gcc-11 from Nova docker image to fix the GLIBCXX issue
2. Use CUDA installation from Nova to find all the necessary CUDA libraries there.  However, `LD_LIBRARY_PATH`, `CUDNN_INCLUDE_DIR`, and `CUDNN_LIBRARY` env variables need to be updated to point there.  It looks like FBGEMM needs to do relocate the wheel similar to what [vision does](https://github.com/pytorch/vision/blob/main/packaging/wheel/relocate.py). What do you think? @atalman @q10 
3. The CUDA aarch64 build is different than CUDA x86 build because it runs on regular arm runner without NVIDIA gpu, so it doesn't have CUDA installation nor nvidia-smi. This leads to this failure https://github.com/pytorch/FBGEMM/actions/runs/12485012915/job/34843246877#step:18:242. Luckily, FBGEMM already has a logic to fix that, which I reuse.

huydhn:
Here is the fix to CUDA aarch64 wheel build failure https://github.com/pytorch/test-infra/pull/6112, it shouldn't have been there in the first place

PR #3531, opened 2024-12-23
Use 3.12 for building docs

huydhn:
I'm switching to 3.12 to build FBGEMM docs for now.  The trigger of the failure is that we now have torch 3.13t as an experimental nightly build.

cc @atalman There seems to be a mix up in how 3.13 and 3.13t are used in this workflow.  This could mean troubles when people try to use torch 3.13 and 3.13t.

### Testing

https://github.com/pytorch/FBGEMM/actions/runs/12474407777/job/34816361009?pr=3531

### DIFF
--- a/.github/workflows/fbgemm_gpu_docs.yml
+++ b/.github/workflows/fbgemm_gpu_docs.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.13" ]
+        python-version: [ "3.12" ]
 
     steps:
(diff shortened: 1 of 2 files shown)


huydhn:
Ah, I think the confusion comes from this step https://github.com/pytorch/FBGEMM/actions/runs/12462565932/job/34783664686#step:13:15, where `conda install -n build_binary -c conda-forge -y numpy` pulls in python 3.13t.  Anyway, I think we can use 3.12 like this to build docs to keep thing simpler.

PR #3799, opened 2025-03-11
Fix fbgemm_gpu hashtable bench on CPU

pnfox:
The hashtable benchmark in split_table_batched_embeddings_benchmark.py has the --use-cpu flag but didn't use it.


pnfox:
Hashtable benchmark from split_table_batched_embeddings_benchmark.py has been removed. Closing PR

PR #3842, opened 2025-03-18
Use enum to select floating point format in FbgemmEmbedding APIs

MatzeB:
Summary:
X-link: https://github.com/pytorch/pytorch/pull/149338

Most FBGemmEmbedding APIs currently feature a `bool is_bf16_out` parameter to differentiate between the float16 and bfloat16 format when the output array has type `uint16_t`.

I am in the process of adding E5M2 and E4M3FN formats for output arrays with type `uint8_t`. Instead of adding another parameter, I would like to change the `bool is_bf16_out` parameter to `enum FloatFormat` to make it easier to add new formats:

```
enum class FloatFormat {
  DEFAULT,
  FLOAT16,
  BFLOAT16,
  FP8_E5M2,
  FP8_E4M3FN,
};
```

Reviewed By: excelle08

Differential Revision: D68046358

PR #3858, opened 2025-03-20
Revert "Use enum to select floating point format in FbgemmEmbedding APIs"

MatzeB:
Reverts pytorch/FBGEMM#3842

I was not able to land the corresponding pytorch commit https://github.com/pytorch/pytorch/pull/149338 so preparing the revert commit now so it is ready when needed.


q10:
This PR can be closed, as the equivalent https://github.com/pytorch/FBGEMM/pull/3865 has landed

PR #3890, opened 2025-03-27
[fbgemm_gpu] Update Nova jobs

q10:
- Update Nova jobs


.github/workflows/build_wheels_genai_linux_aarch64.yml
```diff
@@ -0,0 +1,66 @@
...
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: main
+      with-cuda: disable
+
+  build:
+    if: ${{ github.repository_owner == 'pytorch' }}
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix:
```
huydhn, line 41: Nit: There is only one configuration here so you could consider get rid of the `matrix` and set these parameters directly like the x86 workflow

huydhn (approved):

PR #4113, opened 2025-05-12
introduce kernel for converting e4m3fn kv_cache to e4m3fnuz

bradleyhd:
Summary:
X-link: https://github.com/facebookresearch/FBGEMM/pull/1197

As titled. NV uses e4m3fn for fp8 KV cache, and AMD uses e4m3fnuz. If we want to do HH serving with NV prefill and AMD decode, we need a way to convert between the two.

In practice this can be done with two operations: overwriting -0 (which are valid in e4m3fn but not in uz) with +0, and multiplying the scale qparam by 2, since the exponent bias differs by 1 between the two formats. The scale qparam is packed as the low bits of a __half2 (fp16x2).

Differential Revision: D74502334


ig-raiderler:
This is a well-structured codebase. I appreciate the clean separation of concerns.

PR #4187, opened 2025-05-26
Add feature evict for dram_kv_embedding_cache.

ArronHZG:
# Feature Eviction

## 1. **Background and Design Philosophy of Feature Eviction**

**Background**  
In sparse training scenarios, sample features exhibit significant frequency disparities. Low-frequency features not only contribute minimally to model training but also lead to memory inefficiencies and overfitting. Thus, a feature eviction strategy is necessary to prune unhelpful features and optimize memory usage.

**Design Goals**  
Develop a generic feature eviction mechanism for `dram_kv_embedding_cache` that supports multiple eviction strategies, enabling efficient utilization of memory resources.

**Design Approach**  
- **Strategy Selection**: Implement mainstream eviction strategies such as count-based (LFU) and timestamp-based (LRU) using polymorphism for extensibility.  
- **Trigger Mechanisms**: Support manual triggers, fixed-step intervals, and memory threshold-based triggers for flexible configuration.  
- **Metadata Storage**: Integrate feature statistics (timestamps, counters, usage status) into memory block metadata to avoid additional memory overhead.  
- **Execution Process**: Overlap training and eviction asynchronously by pausing/resuming eviction tasks, minimizing impact on training efficiency.

### 2.1 **Basic Eviction Strategies**

**Strategy Descriptions**  
- **Count-Based Eviction**  
  Evict features based on their occurrence frequency (Count) in training data. Low-frequency features are deemed "useless". Supports decay mechanisms for handling time-varying data distributions (e.g., periodic business patterns).  
  - Increment count (`count += 1`) on each feature update (capped at `2^31-1` to prevent overflow).  
  - Apply decay during eviction and features where `count < threshold`.  

- **Timestamp-Based Eviction**  
  Track the last update timestamp (in seconds) of each feature and those inactive for extended periods.  
  - Update timestamp (`timestamp = current_step` or system time) on each feature update.  
  - Evict features where `current_step - timestamp > ttl`.  

- **Combined Count+Timestamp Eviction (Hybrid Strategy)**  
  Retain features that are both "frequently occurring" and "recently active" by evaluating:  
  `(count < min_count) AND (current_step - timestamp > ttl)`.  
  Thresholds are determined via feature distribution analysis (e.g., count-timestamp 2D histograms).  

- **L2 Norm-Based Eviction (Contribution Strategy)**  
  Measure feature contribution via the L2 norm of their embeddings; features with smaller norms (e.g., near-zero embeddings) are considered less impactful.  
  - Compute L2 norm (`norm = sqrt(sum(emb^2))`) during eviction.  
  - Evict features where `norm < min_norm`.  
  - Computational complexity: O(d) per feature (d = embedding dimension).  

**Structural Design**  
Implement strategies using polymorphism:  
- **Base Class (`FeatureEvict`)**: Manages asynchronous task scheduling, state tracking, and sharded processing.  
- **Subclasses (`CounterBasedEvict`/`TimeBasedEvict`)**: Implement specific eviction logic (e.g., count decay, timestamp comparison).

### 2.2 **Eviction Trigger Mechanisms**  
| Trigger Type       | Implementation                     |  
|--------------------|------------------------------------|  
| Manual Trigger     | Explicit call via `trigger_evict()`|  
| Fixed Step Trigger | Periodic invocation within training loop |  
| Memory Threshold   | Automatic trigger based on memory usage monitoring |

### 2.3 **Metadata Storage for Feature Eviction**  
**Metadata Structure**  
- `MetaHeader` (16 bytes, memory-aligned): Uses bit fields to compress `counter` and `used` fields, supporting all three eviction strategies.  

metaheader

**Storage Layout**  
- `key`: Feature ID (8 bytes).  
- `timestamp`: Last update timestamp (4 bytes, range 0–4,294,967,295).  
- `count`: Feature occurrence count (31 bits, max `2^31-1`).  
- `used`: Memory block occupancy flag (1 bit).

### 3. **Overlap Implementation of Feature Eviction and Training**

**Core Interfaces**  
The `FeatureEvict` class provides these key interfaces for `DramKVEmbeddingCache` to overlap eviction with training:  
- `trigger_evict()`: Initiate asynchronous eviction task.  
- `pause()`: Suspend ongoing eviction.  
- `resume()`: Resume paused eviction.  
- `is_evicting()`: Query eviction task status.  

**Usage Pattern**  
Surround hashtable updates/queries with `pause()` and `resume()` to temporarily halt eviction during critical operations, resuming during data loading, network forward/backward passes.

pipeline

**Asynchronous Optimization Techniques**  
- **Sharded Parallel Processing**: Divide eviction tasks by shard, leveraging the hashtable's thread pool for parallel execution to minimize resource contention.  
- **Fine-Grained Interruption**: Check interrupt flags after processing each memory block, track progress with `block_cursors`, and use atomic variables for rapid pause/resume control.  
- **Lock Optimization**: Employ read-write locks (`wlock`) to ensure thread safety between eviction and training operations.


emlin:
one more question about the eviction policy. can I configure max feature count or max memory threshold while enabling l2 norm based eviction?

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
```diff
@@ -0,0 +1,199 @@
...
+    for (int shard_id = 0; shard_id < num_shards_; ++shard_id) {
+      submit_shard_task(shard_id);
+    }
+    return true;
+  };
+
+  // 暂停淘汰过程，如果有进行中的任务返回true, 没有返回false
+  // 在暂停阶段，判断淘汰是否完成
+  bool pause() {
+    std::lock_guard<std::mutex> lock(mutex_);
```
emlin, line 144: is this lock required?

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
```diff
@@ -0,0 +1,199 @@
...
+
+class CounterBasedEvict : public FeatureEvictBase {
+ public:
+  CounterBasedEvict(folly::CPUThreadPoolExecutor* executor,
+                    SynchronizedShardedMap<int64_t, float*>& kv_store,
+                    float decay_rate,
+                    int threshold)
+      : FeatureEvictBase(executor, kv_store),
+        decay_rate_(decay_rate),
+        threshold_(threshold) {}
```
emlin: so the decay rate and threshold still need to be manually set?

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
```diff
@@ -0,0 +1,199 @@
...
+  std::mutex mutex_;  // 接口锁，保证 public 接口 线程安全
+  int num_shards_;    // 并发任务数
+};
+
+class CounterBasedEvict : public FeatureEvictBase {
+ public:
+  CounterBasedEvict(folly::CPUThreadPoolExecutor* executor,
+                    SynchronizedShardedMap<int64_t, float*>& kv_store,
+                    float decay_rate,
+                    int threshold)
```
emlin: can we have different threshold for different feature offset?

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache_wrapper.h
```diff
@@ -67,7 +87,11 @@ class DramKVEmbeddingCacheWrapper : public torch::jit::CustomClassHolder {
   }
 
   void set(at::Tensor indices, at::Tensor weights, at::Tensor count) {
-    return impl_->set(indices, weights, count);
+    impl_->feature_evict_pause();
+    impl_->set(indices, weights, count);
```
emlin, line 116: to improve performance, we'll enable async set for dram backend similar to ssd here: https://github.com/pytorch/FBGEMM/blob/c51295132814c221de2e4556e47e1a7750bb7910/fbgemm_gpu/src/ssd_split_embeddings_cache/kv_db_table_batched_embeddings.cpp#L141

so we cannot directly start evict right after set. If we want to do this, should call wait_util_filling_work_done inside of eviction

emlin:
no matter which eviction trigger is configured, are we able to start eviction when memory reaches threshold?

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache_wrapper.h
```diff
@@ -26,15 +26,34 @@ class DramKVEmbeddingCacheWrapper : public torch::jit::CustomClassHolder {
       int64_t max_D,
       double uniform_init_lower,
       double uniform_init_upper,
+      int evict_trigger_mode,
+      int evict_trigger_strategy,
+      int64_t trigger_step_interval,
+      uint32_t ttl,
```
emlin: there is already a weight_ttl_in_hours, shall we drop that?

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache_wrapper.h
```diff
@@ -26,15 +26,34 @@ class DramKVEmbeddingCacheWrapper : public torch::jit::CustomClassHolder {
       int64_t max_D,
       double uniform_init_lower,
       double uniform_init_upper,
+      int evict_trigger_mode,
```
emlin: the torch pybind interface does not allow int or int32 type, let's change to int64_t?

emlin:
with this pr, the dram backend build failed, wondering if you can run build and unit test based on the latest branch?

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
ArronHZG: We hope that feature elimination is thread-safe and not affected by the calling method. At present, I understand that all interfaces may not be guaranteed to be safe. For example, trigger may be called by one thread, and pause is called by another thread, so locking is required.

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
ArronHZG: Yeah, there's no good way to set this automatically.

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
ArronHZG: OK

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache_wrapper.h
ArronHZG: OK

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache_wrapper.h
ArronHZG: 1. Yes, we can continue to improve the trigger timing and confirm the all reduce method of size.
2. OK

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache_wrapper.h
ArronHZG: OK

ArronHZG:
Some third-party packages cannot work properly in our environment, so we did not compile
dram_kv_embedding_cache.h
dram_kv_embedding_cache_wrapper.h, but directly rewrote the code. The compilation errors can be shown below, let's take a look.  @emlin

ArronHZG:
> no matter which eviction trigger is configured, are we able to start eviction when memory reaches threshold?

It sounds more reasonable. We can change it.

kausv:
CI signals are also failing cc @ionuthristodorescu 

`
10:16:30 AM: /opt/buildhome/miniconda/envs/build_docs/lib/python3.13/site-packages/fbgemm_gpu/split_table_batched_embeddings_ops_training.py:docstring of fbgemm_gpu.split_table_batched_embeddings_ops_training.SplitTableBatchedEmbeddingBagsCodegen:219: ERROR: Unexpected indentation.
10:16:30 AM: /opt/buildhome/miniconda/envs/build_docs/lib/python3.13/site-packages/fbgemm_gpu/split_table_batched_embeddings_ops_inference.py:docstring of fbgemm_gpu.split_table_batched_embeddings_ops_inference.IntNBitTableBatchedEmbeddingBagsCodegen.split_embedding_weights_with_scale_bias:3: ERROR: Unexpected indentation.
`

emlin:
> CI signals are also failing cc @ionuthristodorescu
> 
> `10:16:30 AM: /opt/buildhome/miniconda/envs/build_docs/lib/python3.13/site-packages/fbgemm_gpu/split_table_batched_embeddings_ops_training.py:docstring of fbgemm_gpu.split_table_batched_embeddings_ops_training.SplitTableBatchedEmbeddingBagsCodegen:219: ERROR: Unexpected indentation. 10:16:30 AM: /opt/buildhome/miniconda/envs/build_docs/lib/python3.13/site-packages/fbgemm_gpu/split_table_batched_embeddings_ops_inference.py:docstring of fbgemm_gpu.split_table_batched_embeddings_ops_inference.IntNBitTableBatchedEmbeddingBagsCodegen.split_embedding_weights_with_scale_bias:3: ERROR: Unexpected indentation.`

let me try locally and see if that's the operator integration issue

emlin:
> > CI signals are also failing cc @ionuthristodorescu
> > `10:16:30 AM: /opt/buildhome/miniconda/envs/build_docs/lib/python3.13/site-packages/fbgemm_gpu/split_table_batched_embeddings_ops_training.py:docstring of fbgemm_gpu.split_table_batched_embeddings_ops_training.SplitTableBatchedEmbeddingBagsCodegen:219: ERROR: Unexpected indentation. 10:16:30 AM: /opt/buildhome/miniconda/envs/build_docs/lib/python3.13/site-packages/fbgemm_gpu/split_table_batched_embeddings_ops_inference.py:docstring of fbgemm_gpu.split_table_batched_embeddings_ops_inference.IntNBitTableBatchedEmbeddingBagsCodegen.split_embedding_weights_with_scale_bias:3: ERROR: Unexpected indentation.`
> 
> let me try locally and see if that's the operator integration issue

Looks like the rebase conflicts have to be addressed before I can import the change. @ArronHZG

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache.h
```diff
@@ -63,26 +99,33 @@ class DramKVEmbeddingCache : public kv_db::EmbeddingKVDB {
...
       double uniform_init_upper,
+      int64_t evict_trigger_mode = 0,
+      int64_t trigger_step_interval = 0,
+      int64_t mem_util_threshold_in_GB = 0.0,
+      int64_t evict_trigger_strategy = 1,
+      const std::optional<at::Tensor>& count_thresholds = std::nullopt,
+      const std::optional<at::Tensor>& ttls_in_hour = std::nullopt,
+      const std::optional<at::Tensor>& count_decay_rates = std::nullopt,
+      const std::optional<at::Tensor>& l2_weight_thresholds = std::nullopt,
+      const std::optional<at::Tensor>& embedding_dims = std::nullopt,
```
emlin: what's the difference between this and table_dims @115?

fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache.h
```diff
@@ -249,28 +327,33 @@ class DramKVEmbeddingCache : public kv_db::EmbeddingKVDB {
...
-                              weights[id_index]
-                                      .template data_ptr<weight_type>() +
-                                  weights[id_index].numel(),
-                              data_ptr);
+                              FixedBlockPool::data_ptr<weight_type>(block);
+                          std::copy(weights[id_index]
+                                        .template data_ptr<weight_type>(),
+                                    weights[id_index]
+                                            .template data_ptr<weight_type>() +
+                                        weights[id_index].numel(),
```
emlin, line 348: from here, seems the weight id is not copied to block, where will  you copy weight id to the block?

fbgemm_gpu/src/dram_kv_embedding_cache/fixed_block_pool.h
```diff
@@ -8,20 +8,122 @@
...
+    // Can be extended with other fields: uint32_t click, etc.
+  };
+
+  // Metadata operations
+
+  // Key operations
+  static uint64_t get_key(const void* block) {
+    return reinterpret_cast<const MetaHeader*>(block)->key;
+  }
+  static void set_key(void* block, uint64_t key) {
```
emlin: seems this method is not called?

fbgemm_gpu/src/dram_kv_embedding_cache/fixed_block_pool.h
```diff
@@ -8,20 +8,122 @@
...
+    std::size_t alignment;
+  };
+
+  // Metadata structure (publicly accessible)
+  // alignas(8) MetaHeader >= sizeof(void*), avoid mempool block too small.
+  struct alignas(8) MetaHeader {  // 16bytes
+    int64_t key;                  // feature key 8bytes
+    uint32_t timestamp;   // 4 bytes，the unit is second, uint32 indicates a
+                          // range of over 120 years
+    uint32_t count : 31;  // only 31 bit is used, max value is 2147483647
```
emlin, line 39: wondering are we able to make the key and count as optional. in the inference side, memory is more restricted, and key/count won't be used anyway

emlin: this can be done in the next step

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
```diff
@@ -0,0 +1,549 @@
...
+      }
+    }
+
+    for (size_t i = 0; i < evicted_counts.size(); ++i) {
+      evict_rates[i] = processed_counts[i] > 0
+                           ? (evicted_counts[i] * 100.0f) / processed_counts[i]
+                           : 0.0f;
+    }
+    DLOG(INFO) << fmt::format(
+        "Shard {} completed: \n"
```
emlin, line 245: this is just one time execution is finished, not the whole shard finished right?
as we'll frequently call the process in every batch, log every time might be too many. like DLOG_EVERY_N(INFO, 1000) can sample logs

fbgemm_gpu/src/dram_kv_embedding_cache/feature_evict.h
```diff
@@ -0,0 +1,549 @@
...
+      block_cursors_[shard_id] = 0;
+      shards_finished_[shard_id]->store(false);
+    }
+    metrics_.reset();
+  }
+
+  void submit_shard_task(int shard_id) {
+    if (shards_finished_[shard_id]->load()) return;
+    futures_.emplace_back(folly::via(executor_).thenValue(
+        [this, shard_id](auto&&) { process_shard(shard_id); }));
```
emlin, line 189: wondering if it's too many submit if we evict in every batch?

ArronHZG:
> for different type of eviction policy, do we share the same metadata header structure? for example, if we choose to use feature l2 norm based eviction, do we still need to reserve 4 bytes for timestamp?

In the L2 elimination method, count and timestamp are expected to take up no space, but the struct contains key and used information, which is at least 16 bytes under 8-byte alignment, and the space seems to be unable to be compressed any further.

emlin:
one more question is how do I get all cached value for one id, including key, meta header and weight + optimizer?
that method is needed to checkpoint including metadata

emlin:
also one question about the alignment.
if use the backend in inference side, the weight size may not be 8 bytes aligned. all these are about memory saving, not urgent immediately, but after we merged this pr, we can think about how compress a bit

PR #4262, opened 2025-06-04
Vectorize load/store for FP8 Quantization

flaviotruzzi:
Summary:
X-link: https://github.com/facebookresearch/FBGEMM/pull/1340

Vectorizing memory access for load and store on FloatToFP8 quantization.

It works well for a lot of shapes, but it is particularly bad for unaligned access.  Fallback to original kernel when column shape is not aligned.

Differential Revision: D75563906


fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -295,27 +401,66 @@ Tensor _float_to_FP8rowwise_gpu_t(const Tensor& input, const bool forward) {
...
+            });
+      } else {
+        // Simple thread block configuration for the scalar kernel
+        // Use 256 threads per block with 32 threads in X dimension (for warp
+        // alignment)
+        const int BLOCK_DIM_X = 32; // Keep warp-aligned for best performance
+        const int BLOCK_DIM_Y = 8; // Balance between Y coverage and registers
+        dim3 blockSize(BLOCK_DIM_X, BLOCK_DIM_Y);
+
+        int num_sms = 80;
```
valentinandrei: This code will run on multiple generations GPUs with different SM count. You can use `at::cuda::getCurrentDeviceProperties()->multiProcessorCount` to get SM count.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -157,6 +157,112 @@ __global__ inline void _compute_FP8_quantize_cuda_kernel(
...
+
+  // Calculate output width
+  const auto ncols_aligned = (ncols + 4 - 1) / 4 * 4;
+  const auto output_columns = ncols_aligned + 2 * sizeof(float);
+
+  // Calculate global row index with 2D thread blocks
+  const int64_t gy = blockIdx.y * blockDim.y + threadIdx.y;
+
+  // Early return if row is out of bounds
+  if (gy >= nrows) {
```
valentinandrei: You may want this check early in order to avoid the extra computations above if the thread exits early.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -157,6 +157,112 @@ __global__ inline void _compute_FP8_quantize_cuda_kernel(
...
+  const int64_t output_row_offset = gy * output_columns;
+
+  // Calculate the position where the scale values are stored
+  const int64_t scale_offset = output_row_offset + ncols_aligned;
+  const float scale_value = reinterpret_cast<float*>(&output[scale_offset])[0];
+
+  // Find aligned section boundaries within this row
+  // We'll use vector operations only for the aligned middle section
+  static constexpr int vec_size = VectorSizeTraits<input_t>::value;
+  const int64_t vector_blocks = ncols / vec_size;
```
valentinandrei, line 222: Maybe it's better to write an epilogue/prologue and have the kernel run for sizes that are not divisible to `vec_size`. Otherwise we are leaving perf on the table. Also on line 201 you compute `ncols_aligned`.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -157,6 +157,112 @@ __global__ inline void _compute_FP8_quantize_cuda_kernel(
...
+  using vec_t = aligned_vector<input_t, vec_size>;
+  using vec_i = aligned_vector<uint8_t, vec_size>;
+
+  // 1. Process four elements at a time using vectorization where possible
+  for (int64_t vec_idx = thread_idx; vec_idx < vector_blocks;
+       vec_idx += thread_stride) {
+    const int64_t col_idx = vec_idx * vec_size;
+
+    // Don't access beyond the valid input data
+    if (col_idx + 3 < ncols) {
```
valentinandrei: I would use a constexpr to indicate vec_size. Also what if `input_t` is an fp16 value? Does adding with 3 still work?

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -157,6 +157,112 @@ __global__ inline void _compute_FP8_quantize_cuda_kernel(
...
+
+#pragma unroll
+      for (int i = 0; i < vec_size; ++i) {
+        output_row->val[i] = float_to_hfp8(
+            to_float(input_row->val[i]) * scale_value, ebit, bias, max_pos);
+      }
+    }
+  }
+
+  // 2. Process any remaining elements (less than 4) with scalar operations
```
valentinandrei: This executes only if ncols is not divisible with the number of vector elements but you call the kernel only for well behaved shapes.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -157,6 +157,112 @@ __global__ inline void _compute_FP8_quantize_cuda_kernel(
...
+  const int64_t vector_blocks = ncols / vec_size;
+
+  // Thread assignments for vectorized section
+  const int64_t thread_idx = blockIdx.x * blockDim.x + threadIdx.x;
+  const int64_t thread_stride = blockDim.x * gridDim.x;
+
+  using vec_t = aligned_vector<input_t, vec_size>;
+  using vec_i = aligned_vector<uint8_t, vec_size>;
+
+  // 1. Process four elements at a time using vectorization where possible
```
valentinandrei: I think you also need a prologue in case the starting address of the inputs is not aligned to vec_size.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -295,27 +401,66 @@ Tensor _float_to_FP8rowwise_gpu_t(const Tensor& input, const bool forward) {
...
+      } else {
+        // Simple thread block configuration for the scalar kernel
+        // Use 256 threads per block with 32 threads in X dimension (for warp
+        // alignment)
+        const int BLOCK_DIM_X = 32; // Keep warp-aligned for best performance
+        const int BLOCK_DIM_Y = 8; // Balance between Y coverage and registers
+        dim3 blockSize(BLOCK_DIM_X, BLOCK_DIM_Y);
+
+        int num_sms = 80;
+        int max_blocks_per_sm = 16;
```
valentinandrei: There is a physical limit of 64 in flight warps per SM. That means you can have maximum 2048 threads. However the thread block size has maximum 1024 threads and you can also get that via `at::cuda::getCurrentDeviceProperties()`. However I wouldn't use `max_blocks_per_sm` because this number depends on the block size.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -295,27 +401,66 @@ Tensor _float_to_FP8rowwise_gpu_t(const Tensor& input, const bool forward) {
...
+        int target_blocks_x = num_sms * max_blocks_per_sm;
```
valentinandrei: Since this is a memory bandwidth bound workflow, you may need less SMs to saturate the memory bandwidth. It's worth experimenting.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: It is implemented below, this kernel does work for all shapes, but due to the extra data that is used as parameters that are embedded in the last 4 bytes of each row, it complicates things a bit because multiple threads could be touching the same memory portions.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: For sure, I missed this when I made vec_size dependent on dtype. I think this should create a problem, but the tests passed, I'll try to add another shape to fbgemm tests to ensure this is covered.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: I can remove this part, I added this to ensure the kernel would work in all cases, when we had trailing elements the perf was not as good, so I just changed to fallback to original kernel.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: I originally thought about calling the function, but thought that the function would incur some extra cost, but didn't test it, I'll test that. I did test A100 with 108SMs, but didn't notice any change on perf.

valentinandrei: It most likely won't since the properties may be already cached and there won't be an extra call to CUDA runtime ... but please check.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
valentinandrei: Hmm, it would be interesting to debug why perf dropped. Compile time is also important so it would be preferable to have a single kernel. I suspect that when dealing with multiple problem sizes, you might increase register count. Can you print the reg count before/after or check with nsight compute?

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -296,18 +347,21 @@ Tensor _float_to_FP8rowwise_gpu_t(const Tensor& input, const bool forward) {
 
     {
       const int blockDim_x =
-          std::min(ncols, static_cast<int64_t>(threads_per_block));
+          std::min(ncols / 4, static_cast<int64_t>(threads_per_block));
```
valentinandrei: Don't you have to use a vector length based on datatype instead of a hardcoded value?

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -117,44 +117,102 @@ __global__ inline void _get_FP8_qparam_cuda_kernel(
...
+  using vec_t = aligned_vector<input_t, vec_size>;
+  using vec_i = aligned_vector<uint8_t, vec_size>;
+
+  const int64_t col_idx = thread_idx * vec_size;
+
+  // Don't access beyond the valid input data
+  if (col_idx + (vec_size - 1) < ncols) {
+    // Load vec_size elements - handle both aligned and unaligned cases
+    // correctly
+    const vec_t* input_row =
```
valentinandrei, line 233: If `input` address is not aligned to the vector size (16 bytes), this line may result in `RuntimeError: CUDA error: misaligned address`.

Here is an example on how the check against this risk is done for `torch.cat`: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/Shape.cu#L357

If `input` is not aligned, that may change the kernel structure because you would need to do element-by-element for the beginning of `input` then do aligned loads throughout the middle section and finish with the remainder element-by-element.

valentinandrei (changes requested):

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: This is a good call, I'll try to create an example. Maybe if I pass a view? Because the automated tests generate tensors of all shapes and they are fine, but we always create new tensors aligned.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: Missed this, we are leaving perf on the table for fp16 and bf16. In this case we will generate more threads that will just exit, and might cause thread divergence.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -157,6 +184,77 @@ __global__ inline void _compute_FP8_quantize_cuda_kernel(
...
+  if (gy >= nrows || (thread_idx * vec_size) >= ncols) {
+    return;
+  }
+
+  int ebit = forward ? 4 : 5;
+  int bias = forward ? 15 : 31;
+  float max_pos = forward ? 0.9375 : 0.875;
+
+  // Calculate output width
+  const auto ncols_aligned = (ncols + 4 - 1) / 4 * 4;
```
valentinandrei, line 211: I'd replace the hardcoded `4` value with vec_elements. This may fail for FP16.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -294,7 +385,10 @@ Tensor _float_to_FP8rowwise_gpu_t(const Tensor& input, const bool forward) {
           });
     }
 
-    {
+    const uintptr_t addr = reinterpret_cast<uintptr_t>(&input);
+    const bool misaligned = addr % 16 > 0;
```
valentinandrei: Replace 16 with sizeof(vector)

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
```diff
@@ -250,13 +348,6 @@ Tensor _float_to_FP8rowwise_gpu_t(const Tensor& input, const bool forward) {
           C10_CUDA_KERNEL_LAUNCH_CHECK();
         });
   } else {
-    // range_tensor is used to store the range for each embedding row.
-    // We save max_pos/max_val(rowwise) as row scale to quantize
-    // unlike INT8, FP8 does not have zero shift
-    // This will guarantee the numerical match but bring some perf
-    // regression.
-    auto range_tensor = at::empty({nrows}, input.options().dtype(at::kFloat));
```
valentinandrei, line 258: Was this just dead code?

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: This col alignment is not related to memory alignment, but to find the position in the input where the parameters for dequantization are. They are in the same place regardless of dtype, and its like that in all functions.

fbgemm_gpu/src/quantize_ops/quantize_fp8_rowwise.cu
flaviotruzzi: Yes.

PR #4342, opened 2025-06-13
add ckpt and restore with feature evict metaheader

lalala-2:
**1.** Call get_block_size to determine the number of bytes each block contains, including the metaheader.
**2.** During checkpointing, call get_kv_with_metaheader_from_storage to retrieve the weights_with_metaheader for specified IDs. The first dimension of the weights_with_metaheader tensor should match the IDs, and the second dimension's number of bytes should match the byte size obtained in step 1. For example, if using uint8 type, the second dimension should match the byte size acquired in step 1.
**3.** During restoration, call set_kv_with_metaheader_to_storage to restore. Only need to provide weights_with_metaheader, as it contains complete information including keys, values, and feature counts.

### DIFF
--- a/fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache.h
+++ b/fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_cache.h
@@ -22,6 +22,7 @@
 #include <thrift/lib/cpp2/protocol/CompactProtocol.h>
 #include <thrift/lib/cpp2/protocol/Serializer.h>
 #include <torch/script.h>
+#include <cstring>
 
 #include "../ssd_split_embeddings_cache/initializer.h"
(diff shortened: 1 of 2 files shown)


lalala-2:
Due to compilation issues with `fbgemm_gpu/src/dram_kv_embedding_cache/dram_kv_embedding_base.h`, we mocked a `SimpleDramKVEmbeddingCache` in `fbgemm_gpu/test/dram_kv_embedding_cache/ckpt_restore_test.cpp` to test the interfaces `get_block_size`, `get_kv_with_metaheader_from_storage`, and `set_kv_with_metaheader_to_storage`.

q10:
Hi @lalala-2 I believe @bobbyliujb is working on this over at https://github.com/pytorch/FBGEMM/pull/4429

PR #4420, opened 2025-07-01
Initialize variables

cyyever:
There are some changes:
1. ensure that variables are initialised by `cppcoreguidelines-init-variables` check.
2. enable "Wunused-variable" and "Wmaybe-uninitialized" warnings.
3. fix all warnings.


cyyever:
@q10 I have added more fixes, need your help to reimport.

cyyever:
@q10 I fixed the OpenMP problem where clang fails by reverting changes. Also note that the test can compile in my environment, with Clang++ 20.

PR #4423, opened 2025-07-01
Use static functions/variables if possible

cyyever:
There are two changes:
1. Marks in-file templates and other inner functions as static; this provides more opportunities to optimise code, i.e. followed by enabling link time optimization. An unused function in test code is removed.
2. Enables `misc-use-internal-linkage` check.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,6 +6,7 @@
 InheritParentConfig: true
 Checks: '
 bugprone-argument-comment,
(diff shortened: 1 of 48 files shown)


q10:
Build appears to fail

https://github.com/pytorch/FBGEMM/actions/runs/16002043321/job/45151950826?pr=4423

cyyever:
@q10 Some changes are reverted.

PR #4448, opened 2025-07-06
Fix uninitialized access

cyyever:
These bugs are real for small values of M, I don't know why the tests passed before, but it is no harm to add initialization with little overhead.


cyyever:
@q10 The current failures could be solved once #4420 is merged. Then I will rebase to see what happens.

PR #4452, opened 2025-07-08
Add interprocedural optimization option

cyyever:
Add IPO option to enable link time optimization, which is disabled by default. AVX files use inline asm code and can't be parsed by lto implementations, so they are excluded. CUDA code is also excluded.

This optimization has reduced libfbgemm.a from 80Mb to 64Mb on my host, using GCC 15.

We have attempts to enable link time optimization on PyTorch, fixing FBGEMM builds is the important prerequisite.


cyyever:
@q10 I cannot find the fail logs, any idea?

q10:
> @q10 I cannot find the fail logs, any idea?

It looks to be ephemeral CI build issues

PR #4481, opened 2025-07-13
Don't use deprecated asmjit templates

cyyever:
Clean up these deprecation warnings before it is possible to build newer asmjit.


cyyever:
@q10 gtest also failed because of deprecated functions, now it is fixed.

cyyever:
@q10 I disable another warning on MacOS CLAPACK. Since FBGEMM doesn't build on Apple Silicon (for no AVX512), I temporarily disable it. Will recheck the decision later when I can build it on M2.

cyyever:
@q10 Let's restore the option but keep the source code changes because I don't have Intel Macbook at hand..

q10:
@cyyever It looks like the changes break the macOS CI builds.

cyyever:
@q10 More reverts

PR #4491, opened 2025-07-15
Fix MSVC build

cyyever:
This PR fixes all warnings I encountered when trying to build FBGEMM on Windows and integrating the latest FBGEMM to PyTorch. Redundant public settings are removed. Other public settings are removed to `fbgemm_generic` because it is relied by other targets.


cyyever:
@q10 I have fixed clang shared builds.

PR #4509, opened 2025-07-16
[fbgemm] FBGEMM ARM CI

q10:


CMakeLists.txt
```diff
@@ -165,33 +169,104 @@ endif()
...
+
+cpp_library(
+  PREFIX
+    fbgemm_avx512
+  TYPE
+    OBJECT
+  SRCS
+    ${FBGEMM_AVX512_SRCS}
+    ${FBGEMM_AVX512_INLINE_SRCS}
+  INCLUDE_DIRS
```
cyyever: `INCLUDE_DIRS` are not required, because the target can link to the dependencies, and their include paths will be available to FBGEMM. Modern CMake targets shouldn't use include paths.

PR #4524, opened 2025-07-18
[FBGEMM] Modularize FBGEMM CMake Target Definitions, pt 3b

q10:


cyyever:
In PT we use the Sanitizer::address and other similar targets for sanitizers, which are more accurate. Therefore should we check the existence of these targets and prefer them?

bench/CMakeLists.txt
```diff
@@ -31,61 +23,87 @@ if (${BLAS_FOUND})
...
   if(${MKL_FOUND})
-    target_include_directories(${BENCHNAME} PRIVATE "${MKL_INCLUDE_DIR}")
-    target_link_libraries(${BENCHNAME} "${MKL_LIBRARIES}")
+    target_include_directories(${BENCHNAME} PRIVATE
+      ${MKL_INCLUDE_DIR})
+
+    target_link_libraries(${BENCHNAME}
+      ${MKL_LIBRARIES})
+
     target_compile_options(${BENCHNAME} PRIVATE
```
cyyever: Should be target_compile_definitions

bench/CMakeLists.txt
```diff
@@ -31,61 +23,87 @@ if (${BLAS_FOUND})
...
+      -DUSE_MKL)
   endif()
+
   if (${BLAS_FOUND})
-    target_compile_options(${BENCHNAME} PRIVATE "-DUSE_BLAS")
-    target_link_libraries(${BENCHNAME} "${BLAS_LIBRARIES}")
+    target_link_libraries(${BENCHNAME}
+      ${BLAS_LIBRARIES})
+
+    target_compile_options(${BENCHNAME} PRIVATE
```
cyyever: also target_compile_definitions

cmake/modules/CppLibrary.cmake
```diff
@@ -102,12 +102,28 @@ function(cpp_library)
             -Wunknown-pragmas
             -Wimplicit-fallthrough
             -Wno-strict-aliasing
-            -Wunused-variable)
+            -Wunused-variable
+            -Wno-c99-extensions
+            -Wno-infinite-recursion
```
cyyever: `-Wno-infinite-recursion` can now be removed due to my last commit.

cmake/modules/CppLibrary.cmake
```diff
@@ -102,12 +102,28 @@ function(cpp_library)
...
+            -Wno-sign-compare
+            -Wno-gnu-zero-variadic-macro-arguments
+            -Wno-unused-parameter
```
cyyever: We can enable `-Wunused-parameter`, the code is clean enough to support it.

cmake/modules/CppLibrary.cmake
```diff
@@ -102,12 +102,28 @@ function(cpp_library)
...
         elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-            list(APPEND lib_cc_flags -Wmaybe-uninitialized)
+            list(APPEND lib_cc_flags
+                -Wmaybe-uninitialized)
+        endif()
+
+        if(CMAKE_COMPILER_IS_GNUCXX AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8.3.0))
+            # Workaround for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80947
+            list(APPEND lib_cc_flags
+                -Wno-attributes)
```
cyyever: Should we really support old g++? Is it for META builds? but how can they build C++20 code?

cyyever:
Some disabled`-Wno-XXX` warnings can actually be used.

cyyever:
One more concern: we use cpuinfo for runtime CPU architecture detection; meanwhile we also use conditional preprocessing inclusion for build time CPU architecture, are the actions conflicting?

cmake/modules/CppLibrary.cmake
q10: It turns out that pytorch workflows, which build fbgemm as part of the build process, uses gcc 9, see https://github.com/pytorch/FBGEMM/pull/4577

PR #4526, opened 2025-07-19
Remove code for old compilers

cyyever:
The C++20 requirement invalidates some old compilers, their branches can be removed.

### DIFF
--- a/include/fbgemm/FbgemmBuild.h
+++ b/include/fbgemm/FbgemmBuild.h
@@ -30,7 +30,7 @@
 #endif
 #define FBGEMM_ENUM_CLASS_API
 #else
-#if __clang__ || __GNUC__ >= 4 || __INTEL_COMPILER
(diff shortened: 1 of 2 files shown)


src/FbgemmSparseDenseVectorInt8Avx512.cc
```diff
@@ -205,14 +196,7 @@ void SparseDenseInt8MVAvx512(
         res = _mm512_add_epi32(res, c_i32_v);
       }
       // Horizontal reduce
-      // _mm512_reduce_add_epi32 is only available for gcc version > 7
-#if __GNUC__ >= 7
```
q10: Removing this ifdef turns out to break some internal downstream package builds, can we revert this change?

cyyever: Reverted

PR #4548, opened 2025-07-23
[fbgemm_gpu] Enable Comm Ops in OSS GenAI

q10:


fbgemm_gpu/experimental/gen_ai/src/comm/car.cpp
```diff
@@ -272,7 +272,13 @@ void nccl_allreduce(
...
       TORCH_CHECK(false, "unsupported type: ", src.scalar_type());
   }
-#if defined(USE_ROCM)
+
+// ncclAllReduceWithBias landed in OSS RCCL only recently, so it won't be
+// available in public ROCm for a while.  As such, its use has to be guarded by
+// a far-off ROCm version check.
+//    https://github.com/ROCm/rccl/pull/1729
+#if (defined(FBGEMM_FBCODE) && defined(USE_ROCM)) || \
+    (!defined(FBGEMM_FBCODE) && defined(USE_ROCM) && (ROCM_VERSION >= 60600))
```
josemonsalve2, line 281: @q10, I believe the ROCM_VERSION guard here should be higher. Maybe 70000? I am on 6.4 right now and `ncclAllReduceWithBias` is still not supported.

q10: Sure, we can update this.  I don't know what is the release timeline for RCCL so I just picked a higher version in arbitrary manner.

PR #4555, opened 2025-07-24
Enable Apple silicon builds

cyyever:
Enable smaller builds on Apple silicon.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -48,13 +48,14 @@ cc_library(
     ],
     deps = [
         ":fbgemm_autovec",
+        ":fbgemm_base",
+        ":fbgemm_headers",
+    ]+ select({
+        ":linux-x86_64": [
         ":fbgemm_avx2",
         ":fbgemm_inline_avx2",
         ":fbgemm_avx512",
-        ":fbgemm_inline_avx512",
-        ":fbgemm_base",
-        ":fbgemm_headers",
-    ],
+        ":fbgemm_inline_avx512"]}),
     linkstatic = 1,
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,7 +266,7 @@ endif()
 # FBGEMM Neon Target
 ################################################################################
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|ARM64")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|ARM64|arm64")
   message(STATUS "Processor is ${CMAKE_SYSTEM_PROCESSOR}; will build Neon target")
 
   get_filelist("get_fbgemm_inline_neon_srcs(msvc=${MSVC_BOOL})" FBGEMM_NEON_SRCS)
@@ -435,7 +435,9 @@ if(FBGEMM_BUILD_TESTS)
 endif()
 
 if(FBGEMM_BUILD_BENCHMARKS)
-  add_subdirectory(bench)
+  if(NOT CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|ARM64|arm64")
+    add_subdirectory(bench)
+  endif()
 endif()
 
 if(FBGEMM_BUILD_DOCS)
--- a/cmake/modules/CxxCompilerSetup.cmake
+++ b/cmake/modules/CxxCompilerSetup.cmake
@@ -40,7 +40,9 @@ BLOCK_PRINT(
 )
 
 # Strip all symbols from the .SO file after building
-add_link_options($<$<CONFIG:RELEASE>:-s>)
+if(NOT MSVC AND NOT APPLE)
+  add_link_options($<$<CONFIG:RELEASE>:-s>)
+endif()
 
 # Enable compile commands to compile_commands.json for debugging
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
--- a/defs.bzl
+++ b/defs.bzl
@@ -27,6 +27,7 @@ def default_compiler_flags():
 def get_fbgemm_base_srcs():
     return [
         "src/GenerateI8Depthwise.cc",
+        "src/GenerateKernelDirectConvU8S8S32ACC32.cc",
         "src/RefImplementations.cc",
         "src/Utils.cc",
     ]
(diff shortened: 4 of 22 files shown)


q10:
@cyyever looks like there are merge conflicts

cyyever:
@q10 Help import again

cyyever:
@q10 It's ready.

cyyever:
@q10 Because I enabled some ARM64 tests, some buggy code has been discovered. Now I have fixed the found errors.

cyyever:
@q10 Now only Bazel builds fail.

q10:
@cyyever we landed https://github.com/pytorch/FBGEMM/pull/4632 earlier today to address the undefined symbols that came up from landing https://github.com/pytorch/FBGEMM/pull/4604.  You may want to rebase this PR on top to leverage the fixes

cyyever:
@q10 Done.

cyyever:
@q10 check aarch macro along without `FBGEMM_FBCODE` macro because bazel always define `FBGEMM_FBCODE`.
`reduceAvx2` has a trivial loop implementation currently so that it can be used for AARCH, despite its name.

q10:
> @q10 check aarch macro along without `FBGEMM_FBCODE` macro because bazel always define `FBGEMM_FBCODE`. `reduceAvx2` has a trivial loop implementation currently so that it can be used for AARCH, despite its name.

We actually do need to check for `FBGEMM_FBCODE`, because the internal builds will break otherwise.  I'm not sure how the internal build system works, but it will complain if it cannot build AVX code even when ARM options are passed to it.

cyyever:
@q10 Is it because 'BUILD.bazel' always build avx targets without checking cpu architecture? I guess you use bazel internally because I can't find BUCK files.

cyyever:
@q10 I have managed to successfully build by Bazel.

q10:
> @q10 Is it because 'BUILD.bazel' always build avx targets without checking cpu architecture? I guess you use bazel internally because I can't find BUCK files.

No, we use BUCK internally, but the BUCK files are not exported to OSS.  I think at some point, there was traction to get FBGEMM building using only Bazel, but since then, it is not actively maintained, as most of our downstream users use CMake to build.

cyyever:
@q10 Bazel builds succeeded and other failure are unrelated. Check if internal builds can work?

q10:
> @q10 has imported this pull request. If you are a Meta employee, you can view this in [D78881059](https://www.internalfb.com/diff/D78881059?dst_version_fbid=743232475076260).

I confirm that it breaks a lot of downstream builds internally.  It might be possible to rectify this long term, but it is a large enough project given the number of downstream builds it is affecting, and we don't have the bandwidth for this at the moment :(.  For now, let's keep with:

```
#ifndef __aarch64__     ==> #if defined(FBGEMM_FBCODE) || !defined(__aarch64__)
#ifdef __aarch64__       ==> #if !defined(FBGEMM_FBCODE) && defined(__aarch64__)
```

which should have the same effect in OSS builds, but will unblock the internal builds

cyyever:
Let me try

q10:
@cyyever The tests appear to hang with ARM when the target type is shared

cyyever:
@q10 It looks like timeout error, the finished 12.6 build also took more than 2h.

cyyever:
@q10 if we build fbgemm_gpu with latest PyTorch, it is possible that compute_70 is not built and the build time is shorten greatly.

q10:
There's still a bunch of errors in the internal builds.  

The first set involves this error signature:

```
fbgemm/src/./DirectConv.h:209:53: error: declaration requires a global destructor [-Werror,-Wglobal-constructors]
  209 | std::mutex DirectConvCodeGenBase<TA, TB, TC, accT>::rtMutex_;
      |      
```
      
The second set involves:

```
ld.lld: error: undefined symbol: fbgemm::initCRegs(asmjit::_abi_1_13::x86::Emitter*, int, int)
>>> referenced by GenerateKernelDirectConvU8S8S32ACC32.cc:286
```

I think it might have to do with the fact that this PR moved GenerateKernelDirectConvU8S8S32ACC32.cc over in defs.bzl (which is shared by both the internal build and OSS build configurations)

The other concern is with the FBGEMM ARM CI, in particular the slowness of `EmbeddingSpMDMTest` when FBGEMM is built in shared library mode (see https://github.com/pytorch/FBGEMM/actions/runs/16789750161/job/47552487008?pr=4555), and whether anything can be done about this...

q10:
In the internal builds, we still see:

```
ld.lld: error: undefined symbol: fbgemm::initCRegs(asmjit::_abi_1_13::x86::Emitter*, int, int)
>>> referenced by GenerateKernelDirectConvU8S8S32ACC32.cc:286
```

and we also see

```
fbgemm/src/UtilsSve.cc:19:6: error: no function template matches function template specialization 'transpose_sve'
   19 | void transpose_sve(
      |      ^
```

cyyever:
@q10 Sorry, I forgot to revert defs.bzl changes.

cyyever:
@q10 Many of the failures are errors from CI scripts. Should we fix them in other PRs before we consider this one?

q10:
> @q10 Many of the failures are errors from CI scripts. Should we fix them in other PRs before we consider this one?

This is being fixed at the moment, so we can safely ignore those errors for now.  However, the FBGEMM ARM CI still has some failures for the shared library case:

```
      Start 11: TransposeTest
11/11 Test #11: TransposeTest ....................***Failed    0.00 sec
/__w/FBGEMM/FBGEMM/build_shared/test/TransposeTest: symbol lookup error: /__w/FBGEMM/FBGEMM/build_shared/libfbgemm.so: undefined symbol: _ZN6fbgemm30requantizeOutputProcessingAvx2ILb1ELb0ELNS_23QuantizationGranularityE2ELb0ELb1EiLb1EEEvPhPKiRKNS_12block_type_tEiiRKNS_22requantizationParams_tIT4_EE
```

q10:
@cyyever we're still seeing:

```
ld.lld: error: undefined symbol: fbgemm::initCRegs(asmjit::_abi_1_13::x86::Emitter*, int, int)
>>> referenced by GenerateKernelDirectConvU8S8S32ACC32.cc:286
```

This is bc src/GenerateKernelDirectConvU8S8S32ACC32.cc has dependency on `initCRegs`, which is defined in GenerateKernel.cc  Moving `src/GenerateKernelDirectConvU8S8S32ACC32.cc` over from `get_fbgemm_generic_srcs` to `get_fbgemm_base_srcs` in this PR appears to be the cause.  Likewise, moving `src/PackWeightsForDirectConv.cc` in defs.bzl is causing undefined symbol errors

cyyever:
Sorry, the history is quite mess，we shouldn't touch defs.bzl in this PR.

PR #4557, opened 2025-07-24
Enable asmjit static builds on Windows

cyyever:
It is possible to always enable static asmjit builds and reduce the number of installed shared libraries.


cthi:
Looks there is some merge conflicts @cyyever

cyyever:
Resolved

PR #4566, opened 2025-07-25
Unify AVX detection

cyyever:


cyyever:
@q10 This PR seems to conflict with #4524. I will resolve them if that PR is merged first.

q10:
> @q10 This PR seems to conflict with #4524. I will resolve them if that PR is merged first.

Ok, sounds good.  Let me try to get #4524 merged soon

cyyever:
@q10 I just found that we shouldn't detect AVX at compile time. Because fbgemm_generic assumes runtime detection, so that if fbgemm_avx{2,512} are disabled in the build environment, there are template instantiation errors, furthermore they may actually be available on user hosts. So let's unconditionally build fbgemm_avx2 and fbgemm_avx512 unless on non-X86 hosts.

cyyever:
The test failures are unrelated.

q10:
> @q10 I just found that we shouldn't detect AVX at compile time. Because fbgemm_generic assumes runtime detection, so that if fbgemm_avx{2,512} are disabled in the build environment, there are template instantiation errors, furthermore they may actually be available on user hosts. So let's unconditionally build fbgemm_avx2 and fbgemm_avx512 unless on non-X86 hosts.

That makes sense, let me update my PR accordingly

q10:
> > @q10 I just found that we shouldn't detect AVX at compile time. Because fbgemm_generic assumes runtime detection, so that if fbgemm_avx{2,512} are disabled in the build environment, there are template instantiation errors, furthermore they may actually be available on user hosts. So let's unconditionally build fbgemm_avx2 and fbgemm_avx512 unless on non-X86 hosts.
> 
> That makes sense, let me update my PR accordingly

#4524 is almost ready to land, after which we can rebase this PR.  Some adjustments had to be made to make sure the changes build in the internal build system

cyyever:
I'm ready to collaborate with you.

cyyever:
Conflicts are resolved.

CMakeLists.txt
```diff
@@ -192,8 +192,13 @@ set(fbgemm_targets)
...
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|AMD64|i386|i686")
-  if(NOT COMPILER_SUPPORTS_AVX2)
-    message(FATAL_ERROR "The current compiler does not support building AVX2 code")
+  set(AVX2_flags)
+  if(CXX_AVX2_FOUND)
+    if(MSVC)
+      set(AVX2_flags -mavx2)
+    else()
+      set(AVX2_flags -mavx2)
```
q10: What's the difference between the two cases here?

cyyever: Typo

cyyever:
@q10 Build failures are fixed.

cyyever:
@q10 Should fix it.

cyyever:
@q10 Unrelated failure

q10:
@cyyever There's a merge conflict with CppLibrary.cmake

Also, it turns out that torch uses clang12 for its builds that include FBGEMM (see https://github.com/pytorch/FBGEMM/pull/4587), so we should make sure the compiler flag removals don't trigger build errors on clang12

cyyever:
@q10 I see, let's remove only c99 flag because C language is not enabled. 'W-deprecated-enum-enum-conversion' was for asmjit, now that is has been updated, clang shouldn't emit warnings as verified by me.

cyyever:
CI outage

q10:
@cyyever just wondering, would it make better sense to define HAS_AVX512 instead of NO_AVX512, since most use cases are `#ifndef NO_AVX512`, which can be awkward to understand?

cyyever:
@q10 I had tried `HAS_AVX512` but found some code already uses `NO_AVX512` (such as `fbgemm_gpu/cmake/Fbgemm.cmake`). I can change it to `HAS_AVX512` again.

cyyever:
@q10 Let's use builtin AVX macros..

cyyever:
@q10 The failures are due to the conflicting build-time AVX512 detection and runtime AVX512 dispatch. Because some AVX512 kernels are generated and compiled at build-time, some parts of  `fbgemmInstructionSet` function has to be masked by `__AVX512F__`. Otherwise, the runtime environment could enable AVX512 dispatching to null kernels.
  I can't reproduce locally and rely CI to discover the mismatch because I have no AVX512 environment.

cyyever:
It looks like the current failures come from other commits...

q10:
> It looks like the current failures come from other commits...

Yes, this was observed while working on #4524.  Were the tests passing when you used `HAS_AVX512` or `NO_AVX512` definitions?

cyyever:
@q10 The previous failing FP16 tests now have passed. These new failed tests passed before, but I have rebased several times after `NO_AVX512`.

cyyever:
@q10 Should I fix them or wait merging?

q10:
> @q10 Should I fix them or wait merging?

@cyyever Yes, if you can help us fix this issue, that would be great.  We have been working on this problem, and could use some help with investigating this issue.

cyyever:
@q10 It occurs on my workstation also, even without AVX512.

cyyever:
@q10 It looks like `isa == inst_set_t::avx2` is used in `isYMM` and incorrectly triggers AVX512 amsjit code.

cyyever:
@q10 There are numerical errors in `GenEmbeddingSpMDMLookup` when the input type is bf16 and output type is float32 combined with ASMJIT avx2 generation, so that some tests in `EmbeddingSpMDMTest` fail. 
However, the tests always pass if the input type is bf16 and output type is float16. Because float16 can be represented without loss as float32, there must be some bugs in the float32 outType case in `GenEmbeddingSpMDMLookup`.

This combination has been disable on Windows and MacOS, it should also be disabled on Linux..

cyyever:
@q10 The current failures are ConvTest, but I can't trigger them locally because they always success. Also note that the failed CI machines can enable AVX512 optimizations, so that the bugs could be there.
A simple fix is disabling them.

cyyever:
@q10 It works.... :sob: , I have resolved the merge conflicts. Hopefully it can be merged ASAP.

cyyever:
@q10 After examining the code, the biggest difference is that FindAVX runs the instructions, now we make it happy if the code can compile, thus reducing the amount of code changes.

cyyever:
@q10 Looks like these failures are unrelated..

PR #4585, opened 2025-07-28
Use C++20 features of std::atomic templates

cyyever:


cyyever:
@q10 Have fixed the merge conflict.

PR #4630, opened 2025-08-01
Fix triton autotuner Config to be compatible with Triton 3.4.x

q10:
Summary: - Fix triton autotuner Config to be compatible with Triton 3.4.x

Differential Revision: D79462743


jerryzh168 (approved):
Thanks!

PR #4655, opened 2025-08-07
Use inline declaration for static members

cyyever:


cyyever:
@q10 This could fix the first internal error.

cyyever:
The failures are due to timeout of `git clone`

PR #4666, opened 2025-08-09
Fix build issue for FP32 OSS and enable tests

omegacoleman:
Hi @q10 , thanks for the great project!

I noticed your previous work in #3568 . After digging into the issue, I believe the build failures are caused by the missing `-masm=intel` flag on various files [here](https://github.com/omegacoleman/FBGEMM/blob/29686bd7085f69b667bdcf1a2fdd2bfff48854e3/CMakeLists.txt#L146) After adding these files, it works like a charm on my machine. The tests and benchmarks have no issues.

If that's the sole problem, the FP32 kernels, tests, and benchmarks can be enabled for OSS, as in this PR.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -42,9 +42,10 @@ cc_library(
     name = "fbgemm",
     visibility = ["//visibility:public"],
     srcs = get_fbgemm_generic_srcs(),
-    hdrs = glob(["src/*.h"]),
+    hdrs = glob(["src/*.h", "src/fp32/*.h"]),
     includes = [
         "src",
+        "src/fp32",
     ],
     deps = [
         ":fbgemm_autovec",
@@ -62,7 +63,7 @@ cc_library(
 cc_library(
     name = "fbgemm_avx2",
     srcs = get_fbgemm_avx2_srcs(),
-    hdrs = glob(["src/*.h"]),
+    hdrs = glob(["src/*.h", "src/fp32/*.h"]),
     copts = [
         "-m64",
         "-mavx2",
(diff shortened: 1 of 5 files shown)


q10:
Hi @omegacoleman thanks for making this PR.  Yes, at the time, I did not have the bandwidth to investigate the cause of the build error, and so I opted to disable it from the build.  

I think the changes here look good, but it looks like it's failing bazel and windows builds at the moment.  Once the builds are all green, we can move ahead to land the PR.

omegacoleman:
> it's failing bazel and windows builds at the moment

I have fixed the bazel issue.

For windows/msvc, intrinsic kernels have to be implemented to enable them. I'm developing a new LLM inference engine, and if the benchmark indicates that packing the weights is worthwhile, even if it's FP32, I may contribute these new kernels myself to support Windows. However, that is a non-trivial task, and it would be better to live in its own PR. For now, I left FP32 disabled on windows.

q10:
@omegacoleman could you rebase this PR on top of latest main?

omegacoleman:
@q10 done that

q10:
@omegacoleman looks like the FP32Test is failing on ARM

omegacoleman:
@q10 I ignored it along with the FP16 test.

The CI didn't produce useful info. I assume that everything involving a NEON kernel (Kleidi*) is causing problems.

I have reproduced the problem on my MacBook M1, and it seems "Unsupported uArch" is being thrown

omegacoleman:
> "Unsupported uArch" is being thrown

The problem roots deeper than I thought. The kernels are fine, but the build system and ISA detection fail.

1. cpuinfo impl, from our dependencies, is in an early stage on arm64. No SVE detection is performed on macOS, and on Linux, it depends on the kernel.
2. What's worse, the env override is missing on arm64
3. On our side, the ISA selector function and the actual kernels are built with different macro definitions.
4. A few misalignments in the build script also create problems.

I finally got the neon kernels to work with a few sheer hacks. But there will be a long way to go to support arm64 properly.

Issue #4679, opened 2025-08-12
Various problems on arm64


omegacoleman:
I'm testing FBGEMM on a Macbook M1. The kernels are fine, but the build system and ISA detection fail. Currently 

1. cpuinfo impl, from our dependencies, is in an early stage on arm64. No SVE detection is performed on macOS, and on Linux, it depends on the kernel.
2. What's worse, the env override is missing on arm64
3. On our side, the ISA selector function and the actual kernels are built with different macro definitions.
4. A few misalignments in the build script also create problems.

This issue is created to track those problems.

PR #4680, opened 2025-08-12
Fix arm64 issues

omegacoleman:
Partly fixes problems in #4679 

1. SVE and NEON are distinct instruction sets. If you have SVE you should not select NEON isa handlers for performance.
2. otoh NEON is alway present on modern aarch64 (as avx2 for x86). So it should be selected if SVE is not present.
3. defs.bzl listed some NEON sources as SVE sources, I corrected them.
4. FBGEMM_ENABLE_KLEIDIAI needs to be defined for generic sources, for getIsaHandler function to know that the NEON handler is present.
5. After that the tests can be enabled, though the performance seems not worth the effort.. currently

..and then it works on my machine™

Depends on #4666 so we don't have to fix FP32 again

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -187,8 +187,6 @@ def get_fbgemm_inline_avx512_srcs(msvc = False, buck = False):
 def get_fbgemm_inline_sve_srcs(msvc = False, buck = False):
     srcs = [
         "src/FbgemmFP16UKernelsSve128.cc",
-        "src/KleidiAIFP16UKernelsNeon.cc",
-        "src/QuantUtilsNeon.cc",
         "src/UtilsSve.cc",
         "src/FbgemmFloat16ConvertSVE.cc",
     ]
@@ -214,7 +212,12 @@ def get_fbgemm_inline_neon_srcs(msvc = False, buck = False):
     intrinsics_srcs = ["src/UtilsNeon.cc"]
 
     # FP16 kernels contain inline assembly and inline assembly syntax for MSVC is different.
-    asm_srcs = ["src/UtilsNeon.cc"]
+    asm_srcs = [
+        "src/UtilsNeon.cc",
+        "src/KleidiAIFP16UKernelsNeon.cc",
+        "src/fp32/KleidiAIFP32UKernelsNeon.cc",
+        "src/QuantUtilsNeon.cc"
+    ]
     if buck:
         return select({
             "DEFAULT": asm_srcs,
--- a/defs.bzl
+++ b/defs.bzl
@@ -222,7 +222,7 @@ def get_fbgemm_inline_neon_srcs(msvc = False, buck = False):
         return select({
             "DEFAULT": asm_srcs,
             "ovr_config//compiler:cl": intrinsics_srcs,
-            "ovr_config//cpu:arm64": intrinsics_srcs,
+            "ovr_config//cpu:arm64": asm_srcs,
         })
     return asm_srcs if not msvc else intrinsics_srcs
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,10 @@ if(FBGEMM_LIBRARY_TYPE STREQUAL STATIC)
 endif()
 list(REMOVE_ITEM fbgemm_generic_defs "FBGEMM_ENABLE_KLEIDIAI")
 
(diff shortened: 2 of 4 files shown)


q10:
@omegacoleman could you rebase on top of latest main?  We had added some changes to https://github.com/pytorch/FBGEMM/pull/4666 to make sure it builds in the internal build system.

omegacoleman:
@q10 Coincidentally just did that.

Did your internal platform test FP16 SVE128 kernels? I got it built and running, but it results in inaccurate calculations:

[sve128-fail-log.txt](https://github.com/user-attachments/files/21746926/sve128-fail-log.txt)

Here's a fraction of log.

The tests would pass if I hide them behind a macro like now and just use NEON ones.

omegacoleman:
> @q10 Coincidentally just did that.
> 
> Did your internal platform test FP16 SVE128 kernels? I got it built and running, but it results in inaccurate calculations:
> 
> [sve128-fail-log.txt](https://github.com/user-attachments/files/21746926/sve128-fail-log.txt)
> 
> Here's a fraction of log.
> 
> The tests would pass if I hide them behind a macro like now and just use NEON ones.

@Nicoshev As the original author, if you have time pls take a look at this. Thanks!

omegacoleman:
Unfortunately, the Graviton 2 AWS arm runner you are using (t2g.2xlarge) has no FP16FML instruction support. Will have to turn off that test as well.

Consider upgrading to Graviton 3, which is armv8.4-a, so it should support fp16fml by default.

q10:
> Unfortunately, the Graviton 2 AWS arm runner you are using (t2g.2xlarge) has no FP16FML instruction support. Will have to turn off that test as well.
> 
> Consider upgrading to Graviton 3, which is armv8.4-a, so it should support fp16fml by default.

We will look into upgrading the instances soon

q10:
> > Unfortunately, the Graviton 2 AWS arm runner you are using (t2g.2xlarge) has no FP16FML instruction support. Will have to turn off that test as well.
> > Consider upgrading to Graviton 3, which is armv8.4-a, so it should support fp16fml by default.
> 
> We will look into upgrading the instances soon

Adding instances in https://github.com/pytorch/FBGEMM/pull/4693

q10:
@omegacoleman we're seeing this error in the internal builds, likely bc we do cross-compilations:

```
ld.lld: error: undefined symbol: kleidiai::gemmkernel_1x1_Neon_fp16_fA0fB0fC0(fbgemm::GemmParams<unsigned short>*)
>>> referenced by FbgemmFP16.cc
```

We can try prefacing the new aarch64 ifdefs in FbgemmFP16.cc with `!defined(FBGEMM_FBCODE)` so that the ifdefs apply only to OSS

q10:
@omegacoleman  https://github.com/pytorch/FBGEMM/pull/4693 has landed, so you might want to rebase

omegacoleman:
> We can try prefacing the new aarch64 ifdefs

@q10 I'm not sure which, as the symbol is only referenced here:
https://github.com/pytorch/FBGEMM/pull/4680/files#diff-f1405a62b2040986b13fcd3ad08972ffaf65840c88407f0d514d9067d8b67428R74 and ifdefs around this is untouched.

If you are building for arm64 then FBGEMM_ENABLE_KLEIDIAI should not be enabled without building NEON sources.

If you are not building for arm64, then don't enable FBGEMM_ENABLE_KLEIDIAI at all.

I have clicked 'Allow edits my maintainers' btw

defs.bzl
```diff
@@ -214,7 +212,12 @@ def get_fbgemm_inline_neon_srcs(msvc = False, buck = False):
     intrinsics_srcs = ["src/UtilsNeon.cc"]
 
     # FP16 kernels contain inline assembly and inline assembly syntax for MSVC is different.
-    asm_srcs = ["src/UtilsNeon.cc"]
+    asm_srcs = [
```
q10, line 215: @omegacoleman I think the issue is here.  In the internal builds (the `if buck` returns true case), when `ovr_config//cpu:arm64` is true, the asm_srcs are not selected, and so we run into `undefined symbol: kleidiai::gemmkernel_1x1_Neon_fp16_fA0fB0fC0`.

omegacoleman: I changed it to select asm_srcs for arm targets

PR #4691, opened 2025-08-13
Fix cblas_gemm_compute for Apple silicon

cyyever:
Properly implement FBGEMM_USE_REF_KERNEL for Apple silicon so that cblas_gemm_compute works.


cyyever:
@q10 fixed

cyyever:
@q10 src/fp32/KleidiAIFP32UKernelsNeon.cc is not chosen in defs.bzl for any target. That is the reason for the linking errors. I have to temporarily disable the tests on ARM64.

PR #4722, opened 2025-08-18
Enable fp32 tests for Windows

cyyever:
src/fp32/FbgemmFP32.cc could work on Windows by commenting out ASM code.


q10:
@cyyever windows builds appear to be failing still :(

PR #4747, opened 2025-08-20
apply unroll in backward

zhiding512:
apply unroll in backward


q10:
@zhiding512 based on the build errors, I think we might need to wrap some of the header includes with ifdef

q10:
@zhiding512 could you also sign the CLA so that we can import the code to our internal repo?

PR #4850, opened 2025-09-10
Bump ASMJIT to 1.18

cyyever:
The author has changes the naming convention to snake_case. I am not sure whether it will break META internal builds..

### DIFF
--- a/src/AsmjitCompat.h
+++ b/src/AsmjitCompat.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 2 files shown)


cyyever:
@q10 We should check interface compatibility with other META projects using older ASMJIT, just like when we bumped it to v1.17.

q10:
@cyyever I think we might need to hold off on this upgrade for the moment.  

I took a look at the internal codebase, and there appears to be multiple large projects that rely on asmjit, and upgrading asmjit here requires upgrading all of those other projects in unison, which is very risky, not to mention that we unfortunately don't have the bandwidth to undertake and verify such a large change at the moment.

Thank you for your understanding.

cyyever:
@q10 Understand it. That case is anticipated.

cyyever:
@q10 I introduced a compatibility layer.

PR #4878, opened 2025-09-16
Add flags to compile ck gemms under pytorch

pragupta:


jeffdaily:
@q10 another rocm-related PR for your review.  thanks.

PR #4925, opened 2025-09-24
backward performance optimization for MI350

liligwu:
bwd performance optimization for ROCm.
Fix numerical issues

### DIFF
--- a/cmake/modules/CppLibrary.cmake
+++ b/cmake/modules/CppLibrary.cmake
@@ -168,18 +168,6 @@ function(cpp_library)
         target_link_libraries(${lib_name} PUBLIC OpenMP::OpenMP_CXX)
     endif()
 
-    if(NOT TARGET TBB::tbb)
-        find_package(TBB QUIET)
-    endif()
-    if(TBB_FOUND)
-        target_link_libraries(${lib_name} PUBLIC TBB::tbb)
-    else()
-        find_library(TBB_LIB NAMES tbb tbb12 HINTS $ENV{CONDA_PREFIX}/lib /usr/lib/x86_64-linux-gnu /usr/local/lib /lib/x86_64-linux-gnu)
-        if(TBB_LIB)
-            target_link_libraries(${lib_name} PUBLIC ${TBB_LIB})
-        endif()
-    endif()
-
     # Add sanitizer options if needed
     if(args_SANITIZER_OPTIONS)
         target_link_options(${lib_name} PUBLIC
--- a/cmake/modules/GpuCppLibrary.cmake
+++ b/cmake/modules/GpuCppLibrary.cmake
@@ -302,18 +302,6 @@ function(gpu_cpp_library)
         list(APPEND library_dependencies ${NVML_LIB_PATH})
     endif()
 
-    if(NOT TARGET TBB::tbb)
-        find_package(TBB QUIET)
-    endif()
-    if(TBB_FOUND)
-        list(APPEND library_dependencies TBB::tbb)
-    else()
-        find_library(TBB_LIB NAMES tbb tbb12 HINTS $ENV{CONDA_PREFIX}/lib /usr/lib/x86_64-linux-gnu /usr/local/lib /lib/x86_64-linux-gnu)
-        if(TBB_LIB)
-            list(APPEND library_dependencies ${TBB_LIB})
-        endif()
-    endif()
-
     # Link against the external libraries as needed
     target_link_libraries(${lib_name} PRIVATE ${library_dependencies})
 


q10:
@liligwu we're seeing

```
OSError: libtbb.so.12: cannot open shared object file: No such file or directory
```

We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH

liligwu:
> @liligwu we're seeing
> 
> ```
> OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> ```
> 
> We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH

Hi @q10 , sorry I missed your message.
I actually have this commit https://github.com/pytorch/FBGEMM/pull/4925/commits/4d2bfddf472907e443203acc185939fabfabe9f6 that links tbb explicitly, it works in your container.  Do you have any suggestions for fixing this issue in CI, please?

BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.

Thank you.

q10:
> > @liligwu we're seeing
> > ```
> > OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH
> 
> Hi @q10 , sorry I missed your message. I actually have this commit [4d2bfdd](https://github.com/pytorch/FBGEMM/commit/4d2bfddf472907e443203acc185939fabfabe9f6) that links tbb explicitly, it works in your container. Do you have any suggestions for fixing this issue in CI, please?
> 
> BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.
> 
> Thank you.

I think this commit only addresses the build step, where we need to link to tbb.  However, for runtime, you might need to do a find in $CONDA_PREFIX from inside the container, and manually update LD_LIBRARY_PATH, or create a symlink (something like https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L383).

liligwu:
> > > @liligwu we're seeing
> > > ```
> > > OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH
> > 
> > 
> > Hi @q10 , sorry I missed your message. I actually have this commit [4d2bfdd](https://github.com/pytorch/FBGEMM/commit/4d2bfddf472907e443203acc185939fabfabe9f6) that links tbb explicitly, it works in your container. Do you have any suggestions for fixing this issue in CI, please?
> > BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.
> > Thank you.
> 
> I think this commit only addresses the build step, where we need to link to tbb. However, for runtime, you might need to do a find in $CONDA_PREFIX from inside the container, and manually update LD_LIBRARY_PATH, or create a symlink (something like
> 
> https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L383
> 
> ).

Hi @q10 , I can see a few line below of your example that links tbb
https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L388

One more thing is that installing tbb may not be sufficient. 
for example, in CentOS we `dnf install -y tbb-devel tbb`

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1042,7 +1059,22 @@ Tensor {{ embedding_cuda_op }}(
 
                     // Compute shared memory size for cta_per_row
                     constexpr auto kCacheAccBytes = sizeof(at::acc_type<cache_t, true>);
-                    int32_t num_cta_per_row_groups = kMaxThreads / kWarpSize;
```
ionuthristodorescu, line 1045: Is this line and the one below common between CUDA and ROCm ? If yes, we should the `{% if rocm %}` guards around them.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1042,7 +1059,22 @@ Tensor {{ embedding_cuda_op }}(
...
+                    int32_t total_L = indices.numel();
```
ionuthristodorescu, line 1066: See above comment - `total_L` seems used only in the ROCm path, so move it under `ifdef USE_ROCM` ?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -970,7 +987,7 @@ Tensor {{ embedding_cuda_op }}(
                 auto num_long_run_ids = at::zeros({1}, indices.options().dtype(at::kInt));
 
                 const bool use_deterministic_algorithms = at::globalContext().deterministicAlgorithms();
-                const int max_segment_length_per_cta = use_deterministic_algorithms ? INT_MAX : 1024;
+                const int max_segment_length_per_cta = use_deterministic_algorithms ? INT_MAX : 4096;
```
ionuthristodorescu: This seems to affect the CUDA regular path as well - please use a `{% if rocm %}` guard to select between 1024 and 4096.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1042,7 +1059,22 @@ Tensor {{ embedding_cuda_op }}(
 
                     // Compute shared memory size for cta_per_row
                     constexpr auto kCacheAccBytes = sizeof(at::acc_type<cache_t, true>);
-                    int32_t num_cta_per_row_groups = kMaxThreads / kWarpSize;
+                    int32_t total_L = indices.numel();
+                    #ifdef USE_ROCM
```
ionuthristodorescu: `USE_ROCM` is a ROCm-specific, could we guard it with a `{% if rocm %}` so it does not bleed into CUDA codegen ?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1053,12 +1085,13 @@ Tensor {{ embedding_cuda_op }}(
...
 
                     const int32_t cta_per_row_grid_size = std::min(
-                        div_round_up(total_unique_indices, kMaxThreads),
+                        div_round_up(total_unique_indices, work_group_size),
                         get_max_thread_blocks_());
 
                     FBGEMM_LAUNCH_KERNEL(
                         backward_cta_per_row_kernel,
                         cta_per_row_grid_size,
+                        // (64, 2)
```
ionuthristodorescu: Do we need this comment ?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1162,7 +1195,18 @@ Tensor {{ embedding_cuda_op }}(
                              kUseVecBlocking>;
 
                     // Compute shared memory size for warp_per_row
-                    int32_t num_warp_per_row_groups = kBackwardMaxThreads / kThreadGroupSize;
+                    #ifdef USE_ROCM
```
ionuthristodorescu: `USE_ROCM` is a ROCm-specific, could we guard it with a `{% if rocm %}` so it does not bleed into CUDA codegen ?

fbgemm_gpu/codegen/training/forward/embedding_forward_split_template.cu
```diff
@@ -454,6 +458,16 @@ batch_index_select_dim0_codegen_forward_cuda(
 
     CUDA_DEVICE_GUARD(dev_weights);
 
+    #ifdef USE_ROCM
```
ionuthristodorescu: `USE_ROCM` is a ROCm-specific, could we guard it with a `{% if rocm %}` so it does not bleed into CUDA codegen ?

fbgemm_gpu/codegen/training/index_select/batch_index_select_dim0_host.cpp
```diff
@@ -341,7 +341,7 @@ class BatchIndexSelectDim0GPUOp
     Tensor grad_dev_weights;
     TORCH_CHECK_EQ(grad_outputs.size(), 1);
 
-    constexpr int32_t max_segment_length_per_warp = 32;
+    constexpr int32_t max_segment_length_per_warp = 16384;
```
ionuthristodorescu: This path seems common with regular CUDA, could we guard it with a `{% if rocm %}` guard to select between 32 and 16384?

fbgemm_gpu/codegen/training/index_select/batch_index_select_dim0_host.cpp
```diff
@@ -656,7 +656,7 @@ class BatchIndexSelectDim0TensorGPUOp
     const auto permute_output_dim_0_1 =
         ctx->saved_data["permute_output_dim_0_1"].toBool();
 
-    constexpr int32_t max_segment_length_per_warp = 32;
```
ionuthristodorescu, line 661: This path seems common with regular CUDA, could we guard it with a `{% if rocm %}` guard to select between 32 and 16384?

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
```diff
@@ -698,6 +698,7 @@ class {{ autograd_func }} :
     TORCH_CHECK(aux_tensor[IDX_LXU_CACHE_LOCATIONS].has_value(), "lxu_cache_locations should have value.");
     const auto lxu_cache_locations = aux_tensor[IDX_LXU_CACHE_LOCATIONS].value();
     const auto is_experimental = aux_bool[IDX_IS_EXPERIMENTAL_TBE];
+    const auto mixed_D = aux_bool[IDX_MIXED_D];
```
ionuthristodorescu: This path seems common with regular CUDA, could we guard it with a `{% if rocm %}` guard to select between 32 and 16384?

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
```diff
@@ -1005,7 +1006,27 @@ static torch::autograd::variable_list backward(
 
 #ifdef USE_ROCM
     constexpr int32_t BT_block_size = 64;
-    constexpr int32_t max_segment_length_per_warp = 64;
+    int32_t max_segment_length_per_warp = 64;
+    // Workaround. Should not be upstreamed in any way.
+    // Redistribute all cta_per_row work to warp_per_row.
+    int32_t total_L = indices.numel();
+    {%- if (not nobag) and
```
ionuthristodorescu: Could we add a `{% if rocm %}` guard around this code, `total_L` is used only in the `USE_ROCM` path and `USE_ROCM` is ROCm specific (so if we do not guard it with `{% if rocm %}` it will be codegen'ed for regular CUDA paths as well).

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
```diff
@@ -210,7 +214,127 @@ __global__ __launch_bounds__(kForwardMaxThreads) void
             )
             {%- endif %}
 
-            for (auto j = 0; j < kWarpSize && l_start + j < L; ++j) {
```
ionuthristodorescu, line 213: Here we should use a `{% if rocm %}` guard to select between rolled / unrolled versions of the loop on regular, non-ROCm paths.

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
kudomcho: By adding a jinja, do you mean only applying {% if rocm %} around the total_L, i.e. @ionuthristodorescu 

```
{% if rocm %}
    int32_t total_L = indices.numel();
{%- endif %}?
```

Or changing from using USE_ROCM to {% if rocm %} entirely?

fbgemm_gpu/codegen/training/index_select/batch_index_select_dim0_host.cpp
kudomcho: The max_segment_length_per_warp passed by host will be modified on embedding_split_host_pt2_autograd_template.cpp later. Reverting max_segment_length_per_warp on here back to 32.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
kudomcho: Applied {% if is_rocm %} onto here

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
kudomcho: Applied {% if is_rocm %} onto here

fbgemm_gpu/codegen/training/index_select/batch_index_select_dim0_host.cpp
kudomcho: The max_segment_length_per_warp passed by host will be modified on embedding_split_host_pt2_autograd_template.cpp later. Reverting max_segment_length_per_warp on here back to 32.

fbgemm_gpu/codegen/training/forward/embedding_forward_split_template.cu
kudomcho: Replaced {% if is_rocm %} onto here

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
kudomcho: Replaced {% if is_rocm %} onto here

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
kudomcho: Removed.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
kudomcho: Replaced {% if is_rocm %} onto here

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
kudomcho: Replaced {% if is_rocm %} onto here

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
kudomcho: Moved int32_t total_L = indices.numel(); doen under {% if is_rocm %}

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
spcyppt: I don't think rocm needs to be guarded here. mixed_D has been passed to all. It was not used in the forward function, so it's just being saved for backward.

Please use `static_cast<bool>(aux_bool[IDX_MIXED_D]);`

then you can replace this [line](https://github.com/pytorch/FBGEMM/blob/04d590e06995f04326eba854599d1daba3501f29/fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp#L814C34-L814C75) to `ctx->saved_data["mixed_D"] = mixed_D`

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
spcyppt: The source code will generate but it shouldn't be compiled. Are we trying to double check here or does it cause any issues, because CUDA path should not compile this. Besides, I think `rocm` is not passed in this file as a global variable, so the condition will always be false, and total_L will never show up in the source code.

fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops_training.py
```diff
@@ -813,7 +813,7 @@ def __init__(  # noqa C901
             assert (
                 self.pooling_mode != PoolingMode.NONE
             ), "Mixed dimension tables only supported for pooling tables."
-
+        self.mixed_D = mixed_D
```
spcyppt, line 823: On this change, I assume mixed_D needs to be accessible as the module parameter? Would it cause any issues if it's not self.mixed_D? Asking this to check if we need to split the PRs for backend (C++ source code and backend codegen) and frontend changes (split_table_batched_embeddings_ops_training.py).

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
spcyppt: This is common.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
spcyppt: I couldn't find the fixes, not sure if it's another PR I missed or it hasn't been pushed. Would you mind sharing the latest version with all the fixes?

kudomcho: The fix regarding this is already on line 1065 on this PR. This display is outdated.

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
kudomcho: I agree with you. Either USE_ROCM or jinja {% if is_rocm %} is fine for us. Please let us know which one we stick with.

fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops_training.py
kudomcho: Optimization on warp_per_row and cta_per_row kernels will not be activated if self.mixed_D is not present.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
kudomcho: {% if rocm %} guards are already applied. Do we need additional changes on here?

spcyppt: Let me take a look and review in the final version so that we're on the same page.

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
spcyppt: On this file, we should stick with `USE_ROCM`. I think your original change is correct. `rocm` is not defined in this file, so jinja will see it as false, and `int32_t total_L = indices.numel();` may not show up in the generated source code. Let me review in the final version.

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
kudomcho: Applied the suggested code onto latest PR

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
kudomcho: Updated and pushed to this PR.

fbgemm_gpu/cmake/tbe_sources.py
```diff
@@ -176,7 +176,6 @@
             "_nobag" if nobag else "",
         )
         for nobag in [
-            True,
```
spcyppt, line 179: no nobag for gen_embedding_backward_split_{}{}_device_kernel_hip.hip?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
```diff
@@ -209,8 +213,131 @@ __global__ __launch_bounds__(kForwardMaxThreads) void
...
+                        grad_indice_weights[indices_start + l_start + j] = grad_indice_weight0;
+                        grad_indice_weights[indices_start + l_start + j+1] = grad_indice_weight1;
+                        grad_indice_weights[indices_start + l_start + j+2] = grad_indice_weight2;
+                        grad_indice_weights[indices_start + l_start + j+3] = grad_indice_weight3;
+                    }
+                }
+            }
+            {%- endif %}
+            for (; j < kWarpSize && l_start + j < L; ++j) {
+            {%- else %} // if is_rocm
```
spcyppt: nit: change to {#-/* if is_rocm*/#}, otherwise it would be misleading in the generated code

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
```diff
@@ -209,8 +213,131 @@ __global__ __launch_bounds__(kForwardMaxThreads) void
...
+
+                    if (threadIdx.x == 0) {
+                        grad_indice_weights[indices_start + l_start + j] = grad_indice_weight0;
+                        grad_indice_weights[indices_start + l_start + j+1] = grad_indice_weight1;
+                        grad_indice_weights[indices_start + l_start + j+2] = grad_indice_weight2;
+                        grad_indice_weights[indices_start + l_start + j+3] = grad_indice_weight3;
+                    }
+                }
+            }
+            {%- endif %}
```
spcyppt: nit: add {#-/*  if not ssd and not dense and not use_vec_blocking and not vbe  */#}

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
```diff
@@ -209,8 +213,131 @@ __global__ __launch_bounds__(kForwardMaxThreads) void
...
+            for (; j < kWarpSize && l_start + j < L; ++j) {
```
spcyppt, line 336: line 337 and 339, does it really make a difference?

Otherwise, line 337-340 could just be `{%- endif %}{#-/*  if is_rocm */#}, since line 339 (i.e., `for (auto j = 0; j < kWarpSize && l_start + j < L; ++j) {` will be common for both rocm and cuda.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_warp_template.cu
```diff
@@ -538,7 +546,7 @@ batch_index_select_dim0_codegen_backward_kernel_warp_per_row
 
 {%- endif %}
 
-{%- if is_rocm and not is_index_select and optimizer == "rowwise_adagrad" and not dense and not is_gwd_kernel and not vbe and not ssd %}
+{%- if is_optimized_hip_kernel_supported_mode %}
```
spcyppt, line 549: No nobag support?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1042,7 +1062,22 @@ Tensor {{ embedding_cuda_op }}(
...
+                            num_cta_per_row_groups = (kMaxThreads/4) / kWarpSize;
+                            work_group_size = (kMaxThreads/4);
+                        }
+                        else{
+                            num_cta_per_row_groups = kMaxThreads / kWarpSize;
+                            work_group_size = kMaxThreads;
+                        }
+                    {%- else %}
+                        int32_t num_cta_per_row_groups = kMaxThreads / kWarpSize;
+                        int32_t work_group_size = kMaxThreads;
```
spcyppt: make work_group_size const for cuda

`const int32_t work_group_size = kMaxThreads;`

fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops_training.py
spcyppt: mixed_D is set to be true by default through the pipeline. In this case, we should change mixed_D to be False and let the frontend pass the correct mixed_D.

q10:
> > > > @liligwu we're seeing
> > > > ```
> > > > OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH
> > > 
> > > 
> > > Hi @q10 , sorry I missed your message. I actually have this commit [4d2bfdd](https://github.com/pytorch/FBGEMM/commit/4d2bfddf472907e443203acc185939fabfabe9f6) that links tbb explicitly, it works in your container. Do you have any suggestions for fixing this issue in CI, please?
> > > BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.
> > > Thank you.
> > 
> > 
> > I think this commit only addresses the build step, where we need to link to tbb. However, for runtime, you might need to do a find in $CONDA_PREFIX from inside the container, and manually update LD_LIBRARY_PATH, or create a symlink (something like
> > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L383
> > 
> > ).
> 
> Hi @q10 , I can see a few line below of your example that links tbb
> 
> https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L388
> 
> One more thing is that installing tbb may not be sufficient. for example, in CentOS we `dnf install -y tbb-devel tbb`

In regards to the TBB issue, I think this might be fixed with PR 5087.  Could you rebase on latest main after this PR 5087 lands?

avbokovoy:
> > > > > @liligwu we're seeing
> > > > > ```
> > > > > OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH
> > > > 
> > > > 
> > > > Hi @q10 , sorry I missed your message. I actually have this commit [4d2bfdd](https://github.com/pytorch/FBGEMM/commit/4d2bfddf472907e443203acc185939fabfabe9f6) that links tbb explicitly, it works in your container. Do you have any suggestions for fixing this issue in CI, please?
> > > > BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.
> > > > Thank you.
> > > 
> > > 
> > > I think this commit only addresses the build step, where we need to link to tbb. However, for runtime, you might need to do a find in $CONDA_PREFIX from inside the container, and manually update LD_LIBRARY_PATH, or create a symlink (something like
> > > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L383
> > > 
> > > ).
> > 
> > 
> > Hi @q10 , I can see a few line below of your example that links tbb
> > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L388
> > 
> > One more thing is that installing tbb may not be sufficient. for example, in CentOS we `dnf install -y tbb-devel tbb`
> 
> In regards to the TBB issue, I think this might be fixed with PR 5087. Could you rebase on latest main after this PR 5087 lands?

Rebase is done

fbgemm_gpu/fbgemm_gpu/split_table_batched_embeddings_ops_training.py
avbokovoy: Done in https://github.com/pytorch/FBGEMM/pull/4925/commits/d2596c712588f23589a5f960a715d1afc604d949

fbgemm_gpu/cmake/tbe_sources.py
avbokovoy: Not now. The nobag support is planned in near future

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
avbokovoy: Done in https://github.com/pytorch/FBGEMM/pull/4925/commits/585300ded18ac828bd183a51a57dba6ff788836d

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
avbokovoy: Done in https://github.com/pytorch/FBGEMM/pull/4925/commits/585300ded18ac828bd183a51a57dba6ff788836d

fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_warp_template.cu
avbokovoy: Not now. The nobag support is planned in near future

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
avbokovoy: Done in https://github.com/pytorch/FBGEMM/pull/4925/commits/e076556674df4879f96b64a433dd7f1812fafee1

fbgemm_gpu/codegen/training/backward/embedding_backward_split_indice_weights_template.cu
avbokovoy: Done in https://github.com/pytorch/FBGEMM/pull/4925/commits/bf143c73992f8f0242fae89a17c783977cf13694. Now it's unified, so CUDA uses it as a main loop while ROCm uses this loop to handle the tailing iterations

spcyppt:
> > > > > > @liligwu we're seeing
> > > > > > ```
> > > > > > OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> > > > > > ```
> > > > > > 
> > > > > > 
> > > > > >     
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >     
> > > > > >   
> > > > > > We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH
> > > > > 
> > > > > 
> > > > > Hi @q10 , sorry I missed your message. I actually have this commit [4d2bfdd](https://github.com/pytorch/FBGEMM/commit/4d2bfddf472907e443203acc185939fabfabe9f6) that links tbb explicitly, it works in your container. Do you have any suggestions for fixing this issue in CI, please?
> > > > > BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.
> > > > > Thank you.
> > > > 
> > > > 
> > > > I think this commit only addresses the build step, where we need to link to tbb. However, for runtime, you might need to do a find in $CONDA_PREFIX from inside the container, and manually update LD_LIBRARY_PATH, or create a symlink (something like
> > > > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L383
> > > > 
> > > > ).
> > > 
> > > 
> > > Hi @q10 , I can see a few line below of your example that links tbb
> > > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L388
> > > 
> > > One more thing is that installing tbb may not be sufficient. for example, in CentOS we `dnf install -y tbb-devel tbb`
> > 
> > 
> > In regards to the TBB issue, I think this might be fixed with PR 5087. Could you rebase on latest main after this PR 5087 lands?
> 
> Rebase is done

@liligwu  @avbokovoy @q10  We still see the tbb errors on CPU CI? https://github.com/pytorch/FBGEMM/actions/runs/19134142334/job/54730556036

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
```diff
@@ -743,6 +743,7 @@ class {{ autograd_func }} :
     TORCH_CHECK(aux_tensor[IDX_LXU_CACHE_LOCATIONS].has_value(), "lxu_cache_locations should have value.");
     const auto lxu_cache_locations = aux_tensor[IDX_LXU_CACHE_LOCATIONS].value();
     const auto is_experimental = aux_bool[IDX_IS_EXPERIMENTAL_TBE];
+    const auto mixed_D = static_cast<bool>(aux_bool[IDX_MIXED_D]);
     {%- endif %}
```
spcyppt, line 747: actually, this is not needed in forward since you use mixed_D in backward. I already fixed this in the internal diff.

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
```diff
@@ -857,7 +858,7 @@ class {{ autograd_func }} :
 
     {%- if not nobag %}
     ctx->saved_data["max_D"] = max_D;
-    ctx->saved_data["mixed_D"] = static_cast<bool>(aux_bool[IDX_MIXED_D]);
+    ctx->saved_data["mixed_D"] = mixed_D;
```
spcyppt, line 861: so this should be `ctx->saved_data["mixed_D"] = static_cast<bool>(aux_bool[IDX_MIXED_D]);`

I already fixed this in the internal diff.

fbgemm_gpu/codegen/training/pt2/embedding_split_host_pt2_autograd_template.cpp
```diff
@@ -1059,7 +1060,25 @@ static torch::autograd::variable_list backward(
 
 #ifdef USE_ROCM
     constexpr int32_t BT_block_size = 64;
-    constexpr int32_t max_segment_length_per_warp = 64;
+    int32_t max_segment_length_per_warp = 64;
+    int32_t total_L = indices.numel();
```
spcyppt, line 1064: this should be moved under line 1072, since is only used in line 1076 which only exists under `{%- if (not nobag) and
           (optimizer == "rowwise_adagrad") and
           (not vbe) and
           (not is_gwd) and
           (not ssd) and
           (not is_index_select) and
           (not dense) %}`

I already fixed it internal diff.

q10:
> > > > > > > @liligwu we're seeing
> > > > > > > ```
> > > > > > > OSError: libtbb.so.12: cannot open shared object file: No such file or directory
> > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > >     
> > > > > > >       
> > > > > > >     
> > > > > > > 
> > > > > > >       
> > > > > > >     
> > > > > > > 
> > > > > > >     
> > > > > > >   
> > > > > > > We already install tbb [here](https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/utils_build.bash#L373), so it might just be an issue of updating the build scripts to put libtbb in the LD_LIBRARY_PATH
> > > > > > 
> > > > > > 
> > > > > > Hi @q10 , sorry I missed your message. I actually have this commit [4d2bfdd](https://github.com/pytorch/FBGEMM/commit/4d2bfddf472907e443203acc185939fabfabe9f6) that links tbb explicitly, it works in your container. Do you have any suggestions for fixing this issue in CI, please?
> > > > > > BTW, we discovered a numerical issue in [986cceb](https://github.com/pytorch/FBGEMM/pull/4925/commits/986cceb660f14d5f26bae57b45682c4357052c5e) and reverted it in [85417b4](https://github.com/pytorch/FBGEMM/pull/4925/commits/85417b4f591b4a99b4e2bd7d5a9056f1da0cc157). It unblocks merging the bwd optimization first.
> > > > > > Thank you.
> > > > > 
> > > > > 
> > > > > I think this commit only addresses the build step, where we need to link to tbb. However, for runtime, you might need to do a find in $CONDA_PREFIX from inside the container, and manually update LD_LIBRARY_PATH, or create a symlink (something like
> > > > > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L383
> > > > > 
> > > > > ).
> > > > 
> > > > 
> > > > Hi @q10 , I can see a few line below of your example that links tbb
> > > > https://github.com/pytorch/FBGEMM/blob/0d49628d239ab9ad68c31fe9db68a6ac6576098a/.github/scripts/utils_build.bash#L388
> > > > 
> > > > One more thing is that installing tbb may not be sufficient. for example, in CentOS we `dnf install -y tbb-devel tbb`
> > > 
> > > 
> > > In regards to the TBB issue, I think this might be fixed with PR 5087. Could you rebase on latest main after this PR 5087 lands?
> > 
> > 
> > Rebase is done
> 
> @liligwu @avbokovoy @q10 We still see the tbb errors on CPU CI? https://github.com/pytorch/FBGEMM/actions/runs/19134142334/job/54730556036

We had missed one case, and this should be addressed with PR 5106. This PR should be rebased on latest main once PR 5106 lands

fbgemm_gpu/include/fbgemm_gpu/utils/cuda_prelude.cuh
```diff
@@ -21,7 +21,9 @@
 #include <ATen/cuda/CUDAGraphsUtils.cuh>
 #endif
 #include <cassert>
-
+#ifdef USE_ROCM
+#include "fbgemm_gpu/rocm/split_embeddings_common.h"
```
spcyppt, line 25: @liligwu @avbokovoy 
This causes problem with internal downstream builds which the code already has `const auto BLOCK_SIZE = 64;` defined.  Adding `split_embeddings_common.h` which also #define BLOCK_SIZE, caused compilation failure.

shbiswas834: Resolved in [c6b0a88](https://github.com/ROCm/FBGEMM/commit/c6b0a8827fafc860f4c4b59db951749e1b0b71df)

q10:
Hmm, it looks like we are still seeing tbb linking issues - https://github.com/pytorch/FBGEMM/actions/runs/19376974534/job/56094185131?pr=4925

Im guessing that we need both the tbb installation and the explicit tbb linking.  @liligwu sorry for the back and forth, can I ask you to add back the tbb linking in the cmake, and rebase to latest commit on `main`?

PR #4977, opened 2025-10-06
Khanin/merged pool emb opt

kudomcho:
implemented pitch size on tensor allocation for better memory alignment.


fbgemm_gpu/test/merge_pooled_embeddings_test.py
```diff
@@ -127,7 +137,14 @@ def test_all_to_one_device(
     ) -> None:
         dst_device = torch.device(f"cuda:{r.randint(0, num_gpus - 1)}")
         with torch.cuda.device(dst_device):
-            inputs = [torch.randn(10, 20) for _ in range(num_inputs)]
+            pitch = True
+            if pitch:
```
q10: @kudomcho This logic is strange, we set pitch and immediately check its value..

kudomcho: @q10 This is to check the assertion allclose in case pitching is enabled. Currently it forces to True to test the all to one for the pitch condition. Any preference on passing the pitch condition on the test argument?

q10: > @q10 This is to check the assertion allclose in case pitching is enabled. Currently it forces to True to test the all to one for the pitch condition. Any preference on passing the pitch condition on the test argument?

Yes, could you make this an argument to the test method, and use hypothesis @given(...) to pass the selection in?

kudomcho: I added use_pitched on the given at test_all_to_one_device and moved the modified the merged_embeddings_benchmark.py to do pitch calculation by adding --use_pitched.

PR #5074, opened 2025-10-31
Several kernel optimization from aiter team

Bernard-Liu:
Bwd performance optimization for ROCm.
Caution:
please know that, this PR depend on this: **https://github.com/pytorch/FBGEMM/pull/4925** One should merge **4925** first, after that one can merge this PR


fbgemm_gpu/codegen/genscript/optimizers.py
```diff
@@ -228,6 +231,38 @@ def rowwise_adagrad() -> Dict[str, Any]:
     multiplier = SHFL_SYNC(multiplier, 0);
     correction = SHFL_SYNC(correction, 0);
     """
+    split_precomputation_preload += """
```
spcyppt: Please ensure the cuda code is not affected?

Bernard-Liu: We added a "split_precomputation_preload" and this is total a new var, CUDA code would not use it. This is an incremental change, none of the existing logic has been modified. So I think Cuda code would not be affected

fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_cta_template.cu
```diff
@@ -625,7 +625,7 @@ batch_index_select_dim0_codegen_backward_kernel_cta_per_row
     codegen/embedding_common_code_generator.py for more details
 */ #}
 
-{{ instantiate_templates(use_subwarp_shuffle=False) }}
+{{ instantiate_templates(use_subwarp_shuffle=True) }}
```
spcyppt, line 629: this seems to affect NVIDIA?

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
```diff
@@ -11,8 +11,42 @@
...
+        value += __shfl_xor(value, 2, 64);
+        value += __shfl_xor(value, 4, 64);
+        value += __shfl_xor(value, 8, 64);
+        value += __shfl_xor(value, 16, 64);
+        value += __shfl_xor(value, 32, 64);
+    }
+    return value;
+}
+
+#define GROUP_REDUCE_ALL_SUM(val, ...) subwarp_reduce_add<kThreadGroupSize>(val)
```
spcyppt, line 54: this also seems to replace warpReduceAllSum on NVIDIA?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_cta_template.cu
Bernard-Liu: fixed, pls refer: https://github.com/pytorch/FBGEMM/pull/5074/commits/9555b3276244c01297f67534aea0657805a9eef3

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
Bernard-Liu: fixed, pls refer: https://github.com/pytorch/FBGEMM/pull/5074/commits/9555b3276244c01297f67534aea0657805a9eef3

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
```diff
@@ -176,4 +214,164 @@ DEVICE_INLINE void {{ mdesc }}_{{ optimizer }}_table_update_kernel(
...
+    {%- for ph_name in args.placeholder_tensor_names %}
+    {%- set ph_type = "{}_ph_t".format(ph_name) %}
+    typename {{ ph_type }},
+    {%- endfor %}
+    int32_t kFixedMaxVecsPerThread,
+    int32_t kThreadGroupSize = kWarpSize,
+    int32_t VEC_WIDTH,
+    bool kUseVecBlocking
+>
+DEVICE_INLINE void {{ mdesc }}_{{ optimizer }}_table_update_kernel(
```
spcyppt, line 235: should this start with hip_ instead of split? Are we overloading the same function? I think AMD start the kernel with `hip_` for rocm customized kernel? `split` We should make sure all follow the same convention.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1078,6 +1239,30 @@ Tensor {{ embedding_cuda_op }}(
                     int32_t num_cta_per_row_groups = kMaxThreads / kWarpSize;
                     const int32_t work_group_size = kMaxThreads;
                     {%- endif %}
+                    {%- if enable_optimized_hip_mixed_D_kernel  %}
+                    auto cta_blockSize = dim3(kThreadGroupSize, num_cta_per_row_groups);
```
spcyppt, line 1170: line 1243 and line 1261 are the same? It does not need to be under if else? You can move it up and remove line 1260-1262 (else part)

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1208,6 +1392,42 @@ Tensor {{ embedding_cuda_op }}(
                     {%- else %}
                         int32_t num_warp_per_row_groups = kBackwardMaxThreads / kThreadGroupSize;
                     {%- endif %}
+                    auto blockSize = dim3(kThreadGroupSize, num_warp_per_row_groups);
+                    {%- if enable_optimized_hip_mixed_D_kernel %}
+                    {%- if vbe %}
+                    if (use_hip_kernel) {
+                    {%- else %}
+                    if (use_hip_kernel && mixed_D) {
```
spcyppt: please add comment that mixed_d optimized kernel is only supported for non-vbe

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
Bernard-Liu: The 'table_update_kernel' func is used in multiple locations. If we add the 'hip_' prefix, we would need to introduce 'if...else...' statements at all those call sites. We believe overloading is a better approach as it doesn't affect the existing logic and avoids making the code more complex.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
Bernard-Liu: modified

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
Bernard-Liu: added

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
```diff
@@ -11,8 +11,46 @@
...
+                                               optimizer == "rowwise_adagrad" and
+                                               not dense and
+                                               not is_index_select and
+                                               not is_gwd_kernel and
+                                               not nobag and
+                                               not ssd %}
+{%- if is_rocm %}
+template<int32_t kThreadGroupSize, typename T>
+DEVICE_INLINE __device__ T subwarp_reduce_add(T value) {
+    static_assert(kThreadGroupSize == 8 || kThreadGroupSize == 16 || kThreadGroupSize == 32 || kThreadGroupSize == 64, "Wavefront size must be 16/32/64");
```
spcyppt, line 24: is kThreadGroupSize == 8 supported? If so the if else below misses 8. If not, the assert is wrong.

yadaish: @spcyppt @Bernard-Liu 
What about using the following warpReduceAllSum? 
https://github.com/pytorch/FBGEMM/blob/eae9f0d69be06d32a75eb4821ce977c55c29a611/fbgemm_gpu/include/fbgemm_gpu/utils/cuda_prelude.cuh#L141-L158

spcyppt: @Bernard-Liu @yadaish sorry, I wasn't sure if the comment was meant for me.

why do we need `subwarp_reduce_add`? `warpReduceAllSum` is the existing implementation. This PR introduces `subwarp_reduce_add` to replace `warpReduceAllSum`.

subwarp_reduce_add is missing implementation for kThreadGroupSize = 8. Assert statement and implementation does not align. So, with the current commit, the value will just return the value without summing the value across warp for kThreadGroupSize = 8, which will lead to incorrect result?

Not sure what does `subwarp_reduce_add` bring to the table - perf? 
The PR should be updated, either to fix subwarp_reduce_add to support  kThreadGroupSize = 8 or remove it.

Please correct me if I misunderstand anything.

liuxuepdf: fixed, pls refer:

https://github.com/ROCm/FBGEMM/commit/e3682f468dbd9baf599fccbaa076e82e5611de90

https://github.com/ROCm/FBGEMM/commit/5cd0f3a114f4907ac316c384d5b1049b021198a1

fbgemm_gpu/codegen/genscript/optimizers.py
spcyppt: I don't see the point of split_precomputation_preload.

From inspecting line by line, Line 234-265 is basically a duplicate of line 201-232 except line 247`new_sum_square_grads += momentum1_val;` ?

can we not duplicate the code? This makes it hard to maintain and error prone (e.g., contributors may only update one blob).

Bernard-Liu: change that by using string template, pls refer: https://github.com/ROCm/FBGEMM/commit/91756a021f68babffaf5aee6fc836b08f8a64cc7

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1092,6 +1239,29 @@ Tensor {{ embedding_cuda_op }}(
                     int32_t num_cta_per_row_groups = kMaxThreads / kWarpSize;
                     const int32_t work_group_size = kMaxThreads;
                     {%- endif %}
+                    auto cta_blockSize = dim3(kThreadGroupSize, num_cta_per_row_groups);
+                    {%- if enable_optimized_hip_mixed_D_kernel  %}
+                    if (max_D <= 128) {
```
spcyppt, line 1172: what about when max_D > 128?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1222,6 +1391,45 @@ Tensor {{ embedding_cuda_op }}(
                     {%- else %}
                         int32_t num_warp_per_row_groups = kBackwardMaxThreads / kThreadGroupSize;
                     {%- endif %}
+                    auto blockSize = dim3(kThreadGroupSize, num_warp_per_row_groups);
+                    {%- if enable_optimized_hip_mixed_D_kernel %}
```
spcyppt, line 1325: could you check the logic from line 1395 - 1402 and improve the comment? based on the comment both should use hip_mixed_d_warp_kernel. 

So for vbe, always use hip_mixed_d_warp_kernel. For non-vbe, only use hip_mixed_d_warp_kernel for when mixed_D? if it's not mixed D and D are in [32, ..., xxx ], use hip_warp kernel, else use regular warp kernel?

It is quite confusing. Can you help elaborate on what to expect?

spcyppt (commented):
can you help add the comment to the logic of how the kernels should get launched and what it's supposed to replace? It gets kind of confusing what cases the kernel support.

Can you add in the summary on how hip optimized kernels are expected to be launched?
For example, hip_warp will replace the normal warp for D = [64, 128, 160, 192, 256, 320],  all tables have the same D and on HBM, and weights/output have the same precision. CTA will launch normally. 

What variants do hip_mixed_d kernel support and what limitation. Is it supposed to replace the normal warp kernel entirely (e.g., either hip_warp or hip_mixed_d_warp) or in what cases do the normal hipified version get launched.

I want to ensure I understand the expectations of the kernels correctly so I can review/edits accordingly. Thank you.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
Bernard-Liu: just use origin kernel: https://github.com/ROCm/FBGEMM/blob/91756a021f68babffaf5aee6fc836b08f8a64cc7/fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu#L1213
for CTA kernel, we only do optimization for case where max_D <= 128 in this PR

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
Bernard-Liu: In this PR, we have made targeted performance optimizations for certain backward kernels. The specific configs where these optimizations take effect can be found here: https://github.com/ROCm/FBGEMM/blob/aiter/mi350_kernel_opt/fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu#L60 . We applied targeted optimizations for kernels under these conditions, while maintaining compatibility with previously existing optimizations. As a result, the Jinja template conditional logic may appear somewhat complex. I have added some comments for clarification — please refer to: https://github.com/ROCm/FBGEMM/commit/2c4602fe2e49ba9e67fc28532bfeb9da7c2faac1

spcyppt: @Bernard-Liu  sorry for not being clear, but that's not what I asked.  I understand the jinja template logic, I was not sure of the use cases.

https://github.com/ROCm/FBGEMM/blob/aiter/mi350_kernel_opt/fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu#L60 determines when what variants the kernel would be created.

The kernel could be created (i.e., exists) but not necessary get used. My question is when it gets used. 

As mentioned in the previous comment, optimized hip_warp kernel has limited support. **Only** when no mixed D, const batch size, D = [64, 128, 160, 192, 256, 320], all weights on HBM, same dtypes of weights that this kernel will get launched. To give some contexts, when this hip_warp kernel got merged, we assumed it supported UVM/UVM caching and mixed dtypes since there was no guards against it. Then it caused many issues and crashes, we spent time investigating just to learn that the kernel did not support these cases. Those additional conditions were added later after the issues already occurred.

So to prevent this from happening, I want to confirm that we have the mutual understanding on what are the cases that the kernel support and have guards against cases it does not. Or should we assume the kernel is supposed to work for all the cases.

Is this hip_mixed_d (in this PR) to replace hip_warp kernel (which does not support vbe and mixed D)? Or it's meant to replace general warp kernel? Does it still has the same limitation as hip_warp kernel. For example, does the kernel work with UVM/UVM caching? does it work when weights and output have different precision (e.g., weights being FP16 and output being FP32, or does it have BF16 support, etc.). What embedding dims it supports?

cc: @liligwu

Bernard-Liu: Thanks for the detailed question. Let me clarify the design:

**Purpose**: hip_mixed_d_warp is a complement/fallback for hip_warp, NOT a replacement. It handles the cases that hip_warp cannot (eg: VBE and mixed_D), while hip_warp continues to handle its supported cases (non-VBE, non-mixed_D, D in [64,128,160,192,256,320]).

**Capability of hip_mixed_d_warp kernel**:
We designed hip_mixed_d_warp to be as general-purpose as possible, aiming to maintain the same level of applicability and compatibility as original kernel.
1. UVM / UVM Caching: Supported. The table_update_kernel overload correctly handles all PlacementType variants (DEVICE, MANAGED, MANAGED_CACHING). WeightRow handles cache-hit vs cache-miss paths correctly. But, we did not have and did not do local unitests for UVM / UVM caching cases, so I suggest to make some tests to verify this point.

2. Mixed precision (e.g., weights FP16, output FP32): Supported. emb_t and grad_t are strictly separated, bridged via Vec4TAcc<cache_t> (float accumulation).

3. BF16: Technically supported (Vec4T has BF16 load/store). 

4. Embedding dims: Any D is supported via dispatched template parameters. For max_D <= 128, we use a specialized (kFixedMaxVecsPerThread=1, kThreadGroupSize=32) configuration.

spcyppt: @Bernard-Liu thank you so much for the clarification!

So just to confirm, hip_mixed_d_warp is to replace regular split_*_warp for mixed embedding dims, right?

Basically:
For mixed D, hip_mixed_d_warp is always launched. 
For same D:
--- if all the conditions are met -> hip_warp kernel
--- else -> regular split_warp kernel.

Please correct me if I misunderstand anything

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -307,6 +315,147 @@ hip_split_embedding{{ ndesc }}_backward_codegen_{{ optimizer }}_{{ wdesc }}{{ vd
...
+    typename cache_t,
+    typename index_t,
+    {%- for ph_name in args.placeholder_tensor_names %}
+    typename {{ ph_name + "_ph_t" }},
+    {%- endfor %}
+    int32_t kFixedMaxVecsPerThread,
+    int32_t kThreadGroupSize,
+    bool kUseVecBlocking>
+__global__ __launch_bounds__(kMaxThreads) void
+hip_mixed_d_split_embedding{{ ndesc }}_backward_codegen_{{ optimizer }}_{{ wdesc }}{{ vdesc }}_kernel_cta_per_row_1(
```
spcyppt: @Bernard-Liu thanks again for the clarification.

You defined `hip_mixed_d_split_*_cta` here, but it is not being used?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
spcyppt: why do we only apply optimization if enable_optimized_hip_mixed_D_kernel is true? Does it not work for general case? I mean I understand for dense and index_select. But for other optimizers (non-rowwise adagrad), global weight decay, ssd and probably even nobag, this optimization can also apply?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
Bernard-Liu: Yes, that is basically the right understanding, with two clarifications:
1. This is true when `enable_optimized_hip_mixed_D_kernel` and `use_hip_kernel` are enabled.
2. In addition to mixed D, the VBE case also routes to `hip_mixed_d_warp` for same D, since `hip_warp` does not support VBE.

spcyppt: Thank you for the additional remarks.

so let me be specific:
```
for case of pooled rowwise adagrad split TBE with no GWD (non-index select, non-dense, non-ssd, non-nobag, non-global weight decay), and runtime flag (use_hip_kernel) is on:

const batch size and mixed D, hip_mixed_d_warp is always launched.
const batch size and same D
--- if all the conditions are met -> hip_warp kernel
--- else -> regular split_warp kernel.

For vbe, hip_mixed_d_warp is always launched.
```

@Bernard-Liu can you confirm?

@liligwu This is means we still do not have the kernel for constant batch size **(non-VBE)** when **embedding dims are the same** but have mixed weight/output dtypes (e.g., weight being fp16 and output being fp32)

Bernard-Liu: I think this  what we designed and hope the kernel could work as this  way,  but we  did not  do complete test for all cases.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
Bernard-Liu: removed it. https://github.com/ROCm/FBGEMM/commit/fa0121c5ad669108ca7742eabd24dbf6c31b0969

PR #5120, opened 2025-11-12
embedding forward optimization for rocm

JaxChen29:
optimization on embedding forward for rocm:

1. apply vec4 on embedding vbe forward kernel instead of vec2
2. use preload to optimize  vbe forward kernel
3. As there are 64 threads in rocm, optimize subwarp in embedding forward v2 kernel when embedding dim is from 32 to 64.


q10:
Hi @JaxChen29 could you rebase this PR on top of latest main?

PR #5137, opened 2025-11-15
Updated asmjit & adapted to its latest changes

kobalicek:
* Updated asmjit, which uses a different coding style
  * Modified code that used SSE versions of some vector instructions instead of AVX versions (conversion, moves, scalar operations)
  * Modified code that relied on Xmm/Ymm/Zmm to simply use Asmjit's x86::Vec operand in places where possible (requires more future work to remove the FBGEMMs Xmm/Ymm/Zmm substitutes)
  * Simplified code paths that were using avx2 and avx512
  * Simplified the use of a JitRuntime - ideally there should be only one around to reduce the number of memory mappings, so now there is only one shared JitRuntime - it doesn't need a lock as it's thread-safe

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ endfunction()
 project(
   fbgemm
   VERSION 1.4.0
-  LANGUAGES CXX)
+  LANGUAGES C CXX)
 
 # Add C++ compiler flag detection
 include(CheckCXXCompilerFlag)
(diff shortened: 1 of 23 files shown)


kobalicek:
Take it or dump it - it was just a little exercise for me to understand how much time is required per project to adapt to the new version of AsmJit. The rest of the changes are related to a correct use of AsmJit and a correct use of instructions in AVX/AVX-512 mode.

kobalicek:
I'm closing this as I'm not interested in this anymore.

PR #5476, opened 2026-03-12
Implement pre-sorting, caching and contigous warp processing in group_index_select backward

avbokovoy:
This PR includes 3 algorithmic optimizations for backward pass on ROCm:
1. Contiguous warp indexing (rows are contiguous, but in case of row length > "warp size", parts of the row is processed in strided manner)
2. Warp-level caching
3. rocprim index pre-sort with threshold and disable merge-sort (it's radix or onesweep for now)

### DIFF
--- a/fbgemm_gpu/src/sparse_ops/sparse_group_index.cu
+++ b/fbgemm_gpu/src/sparse_ops/sparse_group_index.cu
@@ -5,6 +5,10 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
+#include <cassert>
+#include <limits>
+#include <type_traits>
+#include <variant>
 
 #include "common.cuh"
 #ifdef USE_ROCM
@@ -45,6 +49,10 @@ template <
     typename scalar_t,
     bool USE_INDEX_SELECT,
     bool USE_VAR_COLS,
+    bool USE_CONTIGUOUS_WARPS, 
+    bool USE_SORTED_INDICES,
+    bool USE_CACHE,
+    bool USE_PACKED_ROWS,
     int UNROLL_FACTOR,
     int COLS_PER_WARP,
     int LOG_COLS_PER_WARP>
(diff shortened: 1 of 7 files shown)


fbgemm_gpu/src/sparse_ops/sparse_group_index.cu
```diff
@@ -219,49 +331,106 @@ DLL_PUBLIC void group_index_select_or_add_cuda(
...
-                  INVOKE_GROUP_INDEX_SELECT_OR_ADD(true, false);
-                }
-              } else {
-                if (use_var_cols) {
-                  INVOKE_GROUP_INDEX_SELECT_OR_ADD(false, true);
-                } else {
-                  INVOKE_GROUP_INDEX_SELECT_OR_ADD(false, false);
-                }
-              }
+              std::visit(
```
q10: Will this be dispatching many more template instantiations than actually used?  If so, this will blow up binary size

q10:
@avbokovoy could you address the OSS build failures?

avbokovoy:
> @avbokovoy could you address the OSS build failures?

@q10 Fixed in https://github.com/pytorch/FBGEMM/pull/5476/changes/8e901673a36bdb357d8a45d5bab1ae868c46b2a2

fbgemm_gpu/src/sparse_ops/sparse_group_index.cu
```diff
@@ -108,92 +159,149 @@ __launch_bounds__(kMaxThreads) void group_index_select_or_add_2d_kernel(
...
+        } else {
           output[row * num_cols + i] = LDG(&input[idx * num_cols + i]);
+        }
+      } else {
+        if constexpr (USE_CACHE) {
+          const bool member_changed = (last_member_id_for_accum != -1 &&
+                                      member_id != last_member_id_for_accum);
+          // Probably might be merged into following if-else cascade
+          if (member_changed) {
+            flush_cache_accumulator(last_member_output_tile, last_member_num_cols);
```
spcyppt, line 325: @avbokovoy @liligwu  is the approach supposed to improve NVIDIA perf as well? Is the change mainly to address rocm?If so, the change should affect rocm only

avbokovoy: @spcyppt This path is for ROCm only. All the changes are guarded with `constexpr if` and `USE_CONTIGUOUS_WAPRS, USE_SORTED_INDICES, USE_CACHE, USE_PACKED_ROWS` variables, which are set to `false` in case of CUDA build, so this template variants are not going to be compiled for CUDA (that means that only else branches are going to be compiled and executed during run-time, see lines 369-431)

PR #5617, opened 2026-04-11
Remove torch_compiled

cyyever:
This PR removes torch_compiled because recent PyTorch versions can support Python 3.12.

### DIFF
--- a/fbgemm_gpu/test/batched_unary_embeddings_test.py
+++ b/fbgemm_gpu/test/batched_unary_embeddings_test.py
@@ -9,10 +9,8 @@
 
 
 import random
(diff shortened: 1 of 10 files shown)


q10:
We still have to support python 3.10 and 3.11, following torch releases - https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/generate_ci_matrix.py#L286

cyyever:
@q10 Recent PyTorch versions provide well support of python 3.10 and 3.11. torch.compile should work well on the combinations since PyTorch 2.6+. If the internal PyTorch is in the range, that is fine.

PR #5618, opened 2026-04-11
Simplify is_torchdynamo_compiling to direct import from torch.compiler

cyyever:


q10:
I think we might want to keep the existing code for backward compatibility reasons; there's a good chance it might break internal published models.

cyyever:
@q10 it should work if the internal modules all use PyTorch 2.

PR #5621, opened 2026-04-12
Use aligned_unique_ptr in more places to avoid leak

cyyever:


q10:
@cyyever CPU builds appear to be failing

cyyever:
@q10 fixed

PR #5623, opened 2026-04-12
Use double in dequant ref/scalar to match FMA precision

cyyever:
This PR increase the precision of ref implementations so that the bit patterns of ref and real results match. The advantages of this optimisation is that it becomes possible to easily identify silent numerical errors in AVX code paths.


cyyever:
@q10 Windows failures are fixed.

PR #5634, opened 2026-04-15
Add bf16 avx2 8bit dequant and Nbit dequant

cyyever:


q10:
@cyyever looks like the CPU huilds are failing

cyyever:
@q10 It is likely that this PR depends on https://github.com/pytorch/FBGEMM/pull/5635 because FBGEMM_CODE affects runtime instruction selection. I will rebase once it is merged..

cyyever:
@q10 rebased it

q10:
@cyyever looks like there are still build failures

cyyever:
@q10 I have interesting findings

cyyever:
@q10 I have relaxed the comparison error bounds

src/QuantUtils.cc
```diff
@@ -792,14 +793,22 @@ void FusedNBitRowwiseQuantizedSBHalfToFloatOrHalfRef(
...
+  requires(!is_uint16_t_of_type_bf16 || !std::is_same_v<OutputType, float>)
 void FusedNBitRowwiseQuantizedSBHalfToFloatOrHalf(
     int bit_rate,
     const uint8_t* input,
     size_t input_rows,
     int input_columns,
     OutputType* output,
     bool scale_bias_last [[maybe_unused]]) {
+  auto ref_fallback = [&]() {
+    FusedNBitRowwiseQuantizedSBHalfToFloatOrHalfRef<
```
q10, line 806: Bug: SVE/Neon N-bit dequant produces fp16 instead of bf16 when is_uint16_t_of_type_bf16=true.

The is_uint16_t_of_type_bf16 template parameter added by this diff is NOT forwarded to the FusedNBitRowwiseQuantizedSBHalfToFloatOrHalfNeon calls for bit_rates 2, 4, and 8. The Neon function (QuantUtilsNeon.cc:672) unconditionally converts to fp16 via vcvt_f16_f32. If called with <float16, true> on an SVE platform, the output buffer will contain fp16-formatted bits where the caller expects bf16, producing incorrect numerical results.

src/EmbeddingSpMDM.cc
```diff
@@ -393,10 +394,13 @@ GenEmbeddingSpMDMLookup<
```
q10, line 393: Performance concern: JIT unroll factor reduced by 1 for bf16 output.

The new rounding code reserves 2 vector registers (bf16_bias_vreg + bf16_tmp_vreg) instead of the old 1 (ones_vreg), decrementing unroll_factor by an additional 1 in bf16 output mode. The unroll factor directly controls how many embedding elements are processed per loop iteration. This may cause a measurable throughput regression for embedding lookups with bf16 output on hot paths. Consider benchmarking before landing.

src/EmbeddingSpMDM.cc
```diff
@@ -868,17 +878,26 @@ GenEmbeddingSpMDMLookup<
                   if (is_fp16_out) {
```
q10, line 868: Subtle correctness note: masked AVX-512 rounding relies on stale bf16_tmp_vreg lanes being masked out.

All operations on bf16_tmp_vreg here are under mask k(1). For masked-off lanes, bf16_tmp_vreg retains its value from a previous iteration. This is safe because the final vpaddd into out_vreg is also masked, so stale values don't propagate. But the reasoning is non-trivial — consider adding a brief comment noting this invariant for future readers.

src/RefImplementations.cc
```diff
@@ -109,8 +109,10 @@ void Float16ToFloat_ref(const float16* src, float* dst, size_t size) {
 
 void FloatToBfloat16_ref(const float* src, bfloat16* dst, size_t size) {
   for (size_t i = 0; i < size; i++) {
-    // Add 2^15 and right shift 16 to do round-nearest
-    dst[i] = (*reinterpret_cast<const uint32_t*>(src + i) + (1 << 15)) >> 16;
+    // Round to nearest, ties to even (IEEE 754)
```
q10: Pre-existing nit: reinterpret_cast instead of std::bit_cast.

This line uses *reinterpret_cast<const uint32_t*>(src + i) which is technically a strict aliasing violation in C++. The scalar version in FloatConversion.h correctly uses std::bit_cast<uint32_t>(src). Consider aligning this for consistency. (Pre-existing, not introduced by this diff.)

src/QuantUtils.cc
cyyever: fixed

src/EmbeddingSpMDM.cc
cyyever: excluded from this PR

src/RefImplementations.cc
cyyever: excluded from this PR

src/EmbeddingSpMDM.cc
cyyever: excluded from this PR

PR #5648, opened 2026-04-16
Fix bf16 rounding to IEEE 754 ties-to-even

cyyever:
Fix fp32→bf16 rounding to round-to-nearest, ties-to-even everywhere: scalar, AVX2, AVX512, SVE, and JIT kernels (previously halves rounded away from zero, differing per backend). aarch64 BFCVT is dispatched at runtime via cpuinfo_has_arm_bf16().

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,12 +156,18 @@ if(NOT MSVC)
 endif()
 
 set(fbgemm_arm_defs)
+set(fbgemm_arm_bf16_flags)
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|ARM64|arm64")
   set(fbgemm_arm_defs
       FBGEMM_FP16_FALLBACK_TO_REF_KERNEL
       FBGEMM_FP32_FALLBACK_TO_REF_KERNEL)
   if(NOT APPLE)
     list(APPEND fbgemm_arm_defs FBGEMM_ENABLE_KLEIDIAI)
+  endif()
+  check_cxx_compiler_flag("-march=armv8.6-a+bf16+fp16fml" COMPILER_SUPPORTS_BF16)
+  if(COMPILER_SUPPORTS_BF16)
+    set(fbgemm_arm_bf16_flags "-march=armv8.6-a+bf16+fp16fml")
+    message(STATUS "Enabling hardware BF16 (BFCVT) for aarch64")
   endif()
 endif()
 
@@ -196,6 +202,8 @@ cpp_library(
     $<BUILD_INTERFACE:${FBGEMM_SOURCE_DIR}/bench>
   MSVC_FLAGS
     /wd4717 /bigobj
+  CC_FLAGS
+    ${fbgemm_arm_bf16_flags}
   SANITIZER_OPTIONS
     "${FBGEMM_USE_SANITIZER}"
   DEFINITIONS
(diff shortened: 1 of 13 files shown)


q10:
Could you provide some summary of the motivations for this change, and how it would impact perf and correctness on (on both ARM and x86)?  It's highly likely that a decision was made a long time ago internally with regards to the rounding technique, and the decision was made with tradeoffs relevant to internal use cases in mind.  Also, could you also provide microbenchmark reproducers with perf numbers?

cyyever:
@q10 We change the rounding mode to ties-to-even because it is IEEE 754 default behaviour, also note that CUDA implementation, PyTorch, NumPy use ties-to-even, FBGemm should be consistent with these libraries to avoid silent numerical precision issues. For example, the following snippet confirms the rounding mode. 
``` 
import torch
import numpy as np


def tie_fp32(hi: int) -> float:
    """fp32 with upper 16 bits = hi, low 16 bits = 0x8000 (a rounding tie)."""
    return np.uint32((hi << 16) | 0x8000).view(np.float32).item()


def torch_bf16_bits(f):
    return torch.tensor([f]).to(torch.bfloat16).view(torch.uint16).item()


# Pairs of adjacent ties: first has even keep-bit (should round down),
# second has odd keep-bit (should round up). Both go to the even neighbor.
print(f"{'fp32 hi':>8}  {'keep LSB':>8}  {'expected':>8}   {'torch':>6}")
for hi in [0x3F80, 0x3F81, 0x4000, 0x4001, 0xBF80, 0xBF81]:
    f = tie_fp32(hi)
    expected = hi if (hi & 1) == 0 else hi + 1  # ties-to-even
    parity = "even" if (hi & 1) == 0 else "odd"
    print(f"  0x{hi:04x}  {parity:>8}    0x{expected:04x}   0x{torch_bf16_bits(f):04x}")
    assert torch_bf16_bits(f) == expected
```
A simple bench generated by LLM for AVX2 reported that
```
  ┌────────────────────────┬───────┬───────┬────────┐                                                          
  │      Workload (N)      │ main  │  PR   │    Δ   │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                          
  │ 256   scalar           │ 0.057 │ 0.096 │  +69 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                          
  │ 256   AVX2 bulk        │ 0.049 │ 0.084 │  +72 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                          
  │ 256   AVX2 asmjit-tail │ 0.088 │ 0.109 │  +24 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                          
  │ 4 K   scalar           │ 0.060 │ 0.099 │  +64 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                          
  │ 4 K   AVX2 bulk        │ 0.055 │ 0.087 │  +59 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                          
  │ 4 K   AVX2 asmjit-tail │ 0.087 │ 0.113 │  +30 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                  
  │ 64 K  scalar           │ 0.074 │ 0.093 │  +26 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                        
  │ 64 K  AVX2 bulk        │ 0.066 │ 0.089 │  +34 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                        
  │ 64 K  AVX2 asmjit-tail │ 0.080 │ 0.104 │  +31 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                        
  │ 1 M   scalar           │ 0.094 │ 0.095 │   +1 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                        
  │ 1 M   AVX2 bulk        │ 0.093 │ 0.094 │   +0 % │                                                          
  ├────────────────────────┼───────┼───────┼────────┤                                                        
  │ 1 M   AVX2 asmjit-tail │ 0.093 │ 0.104 │  +12 % │                                                          
  └────────────────────────┴───────┴───────┴────────┘                                                                
                                                            
```
because the new rounding mode requires 2 times more instructions than the old mode.

PR #5652, opened 2026-04-17
Fix OOB read in _get_padding_value_kernel

cyyever:


q10:
@cyyever can you add a unit test to exercise this case?

cyyever:
@q10 existing tests has covered the normal case, let me add some edge cases.

PR #5825, opened 2026-06-03
Fix backward regression in PR #5712

avbokovoy:
Guarded short segments cases to skip optimized kernel launch. New warp_per_row kernel dispatch logic (`FBGEMM_TBE_ROCM_HIP_BACKWARD_KERNEL=1`):

```
if  (vbe || mixed_D) && num_unique_prev > 0 && total_L <= 2*num_unique_prev
  → launch hip_mixed_d_warp (introduced in the PR)
else !mixed_D && D in [64, 128, 160, 192, 256, 320] && weights_dtype == output_dtype && weights_on_HBM
  → launch hip_warp (introduced previously - no change for this PR)
else 
  → regular split_warp kernel
```

### DIFF
--- a/fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_warp_template.cu
+++ b/fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_warp_template.cu
@@ -41,6 +41,14 @@
                                                  not vbe and
                                                  not ssd %}
 
+{%- set enable_optimized_hip_mixed_D_kernel  = is_rocm and
+                                               optimizer == "rowwise_adagrad" and
+                                               not dense and
+                                               not is_index_select and
+                                               not is_gwd_kernel and
+                                               not nobag and 
+                                               not ssd %}
+
 #include "fbgemm_gpu/embedding_backward_template_helpers.cuh"
 #include "fbgemm_gpu/utils/tensor_accessor_builder.h"
 #include "fbgemm_gpu/split_embeddings_utils.cuh"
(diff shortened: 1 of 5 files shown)


spcyppt:
could you help clarify what cases will use this hip_mixed_d kernel? Please also put it in the PR summary.

From previous discussion with Bernard, this is my understanding:
```
if vbe
→ launch hip_mixed_d_warp kernel (introduced in the PR)
else
if mixed D
→  launch hip_mixed_d_warp (introduced in the PR)
else
if D in [64, 128, 160, 192, 256, 320] && weights_dtype == output_dtype && weights_on_HBM
→ launch hip_warp kernel (introduced previously - no change for this PR)
else
→ regular split_warp kernel
```
 
 could you modify the above based on the new push?

cc: @avbokovoy @liligwu

avbokovoy:
@spcyppt The dispatch logic after this PR should look like this:
```
if  (vbe || mixed_D) && num_unique_prev > 0 && total_L <= 2*num_unique_prev
  → launch hip_mixed_d_warp (introduced in the PR)
else !mixed_D && D in [64, 128, 160, 192, 256, 320] && weights_dtype == output_dtype && weights_on_HBM
  → launch hip_warp (introduced previously - no change for this PR)
else 
  → regular split_warp kernel
```
Let me know if you have other questions

cc: @liligwu

spcyppt:
@avbokovoy could you rebase this? There's a conflict and I could not import the PR. I don't have permission to resolve conflicts.

avbokovoy:
> @avbokovoy could you rebase this? There's a conflict and I could not import the PR. I don't have permission to resolve conflicts.

@spcyppt Rebase is done. Apologies for the delay

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -1228,6 +1379,58 @@ Tensor {{ embedding_cuda_op }}(
...
+                                grad_t,
+                                cache_t,
+                                index_t,
+                                {%- for ph_name in args.placeholder_tensor_names %}
+                                {{ ph_name + "_ph_t" }},
+                                {%- endfor %}
+                                1,
+                                32,
+                                false>;
+                            blockSize = dim3(32, (kBackwardMaxThreads / 2) / 32);
```
spcyppt:
```
num_warp_per_row_groups = (kBackwardMaxThreads / 2) / 32;
blockSize = dim3(32, num_warp_per_row_groups);
```

both launch the same block dim3(32, 8). But this also writes back `num_warp_per_row_groups` = 8, so the downstream grid-size (line ~1447) and smem calc use 8 too; 

Without this change, this leaves `num_warp_per_row_groups` at its `total_L/total_B` value (e.g. 4), so the grid over-provisions in that branch.

I confirmed this with the benchmark, the perfs are mostly the same, except
one case T=1 D=128 B=131072 that without my fix regressed 11% (total cuda time).

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -789,6 +879,31 @@ Tensor {{ embedding_cuda_op }}(
...
+                at::TensorOptions()
+                    .dtype(at::kInt)
+                    .device(at::kCPU)
+                    .pinned_memory(true));
+        }
+        // Schedule async D2H copy for the next backward call.
+        entry.pinned_buf.copy_(
+            sorted_linear_indices_num_runs.slice(0, 0, 1),
+            /*non_blocking=*/true);
+    }
```
spcyppt: The copy_ into entry.pinned_buf is non-blocking (async D2H), and on the next backward call the value is read via entry.pinned_buf.data_ptr<int32_t>()[0] with no explicit stream synchronization before the read. Without a cudaStreamSynchronize (or equivalent event wait) covering the prior stream that scheduled the copy, the host may observe a stale or partially-updated value. Even though the value only gates a dispatch heuristic, an incorrect num_unique_prev can cause the optimized kernel to be launched on inputs it wasn't designed for, or vice versa?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
```diff
@@ -757,6 +834,19 @@ Tensor {{ embedding_cuda_op }}(
...
     const int used_shared_bytes = used_shared_kb << 10;
 
+#ifdef USE_ROCM
+    {%- if enable_optimized_hip_mixed_D_kernel %}
+    const bool use_hip_kernel_flag = fbgemm_gpu::config::is_feature_enabled(
+        fbgemm_gpu::config::FeatureGateName::TBE_ROCM_HIP_BACKWARD_KERNEL);
+    // Declared here so the dispatch lambdas below can capture it.
+    // Populated from the pinned buffer after transpose_embedding_input().
+    struct NumUniqueEntry { at::Tensor pinned_buf; };
+    static thread_local std::unordered_map<const void*, NumUniqueEntry>
```
spcyppt: s_num_unique_cache is a static thread_local std::unordered_map<const void*, NumUniqueEntry> keyed by hash_size_cumsum.data_ptr(). Pointer values from freed tensor allocations can be reused by the caching allocator for entirely different tables, causing a subsequent backward on a different table to read a stale num_unique_prev and take the wrong dispatch branch. Additionally, entries are never evicted, so the map grows unbounded across long-running processes/threads. Consider a more stable key (e.g., a tensor identity/version tag) plus a size cap or LRU eviction.

fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_warp_template.cu
```diff
@@ -341,6 +349,307 @@ batch_index_select_dim0_codegen_backward_kernel_warp_per_row(
...
+            {%- for tensor in args.split_tensors %}
+            const auto {{ tensor }}_placement = SUBWARP_SHFL_SYNC(s_{{ tensor }}_placement, i);
+            const int64_t {{ tensor }}_offset = SUBWARP_SHFL_SYNC(s_{{ tensor }}_offset, i);
+            {{ args.split_tensor_types[tensor] }} {{tensor}}_val = SUBWARP_SHFL_SYNC(s_{{ tensor }}_val, i);
+            {%- endfor %}
+
+            // const int64_t momentum1_offset = SHFL_SYNC(s_momentum1_offset, i);
+            // const auto momentum1_placement = static_cast<PlacementType>(SHFL_SYNC(s_momentum1_placement, i));
+            // auto momentum1 = reinterpret_cast<at::acc_type<cache_t, true>*>(SHFL_SYNC(reinterpret_cast<uint64_t>(s_momentum1), i));
+            // auto momentum1_val = momentum1[idx];
```
spcyppt: can you confirm that these can be deleted?

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
```diff
@@ -212,4 +219,164 @@ DEVICE_INLINE void {{ mdesc }}_{{ optimizer }}_table_update_kernel(
...
+        {%- else %}
+        if ({{ locs_or_addrs_idx }} != kCacheLocationMissing) {
+          cache_weights = &lxu_cache_weights[{{ locs_or_addrs_idx }}][0];
+        }
+        {%- endif %}
+    }
+    {%- for tensor in args.split_tensors %}
+    {{ args.split_tensor_types[tensor] }}* __restrict__ {{ tensor }};
+    // const auto {{ tensor }}_placement = static_cast<PlacementType>({{ tensor }}_placements[t]);
+    // const int64_t {{ tensor }}_offset = {{ tensor }}_offsets[t];
```
spcyppt: here as well, are these meant to be deleted?

fbgemm_gpu/codegen/training/backward/embedding_backward_split_kernel_warp_template.cu
avbokovoy: Yes. Deleted in https://github.com/pytorch/FBGEMM/pull/5825/commits/8dca8cf1471b191b22da52146517c12b1fa37411

fbgemm_gpu/codegen/training/optimizer/embedding_optimizer_split_device_kernel_template.cuh
avbokovoy: Deleted in https://github.com/pytorch/FBGEMM/pull/5825/commits/8dca8cf1471b191b22da52146517c12b1fa37411

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
avbokovoy: The whole heuristic is now moved into the kernel to avoid redundant and unreliable memory transfers in https://github.com/pytorch/FBGEMM/pull/5825/commits/95f0ad70edab1052a06b74e83cdec3b53b85e536

fbgemm_gpu/codegen/training/backward/embedding_backward_split_template.cu
avbokovoy: The whole heuristic is now moved into the kernel to avoid redundant and unreliable memory transfers in https://github.com/pytorch/FBGEMM/pull/5825/commits/95f0ad70edab1052a06b74e83cdec3b53b85e536

PR #5952, opened 2026-06-25
Replace TBE type-dispatch macros with C++20 generic-lambda functions

cyyever:
Converts the four TBE type-dispatch macros into namespaced C++20 template functions that take a generic lambda with an explicit template-parameter list, instead of relying on macro-injected `using` aliases.

## Changes
- **`dispatch_macros.h`** — replace `DISPATCH_EMB_CACHE_TYPES`, `DISPATCH_EMB_CACHE_OUTPUT_TYPES`, `DISPATCH_OUTPUT_TYPES`, and `DISPATCH_EMB_GRAD_CACHE_TYPES` with `fbgemm_gpu::dispatch_emb_cache_types`, `dispatch_emb_cache_output_types`, `dispatch_output_types`, and `dispatch_emb_grad_cache_types`.
- **Call sites (16 files across `codegen/` and `src/split_embeddings_cache/`)** — `[&] { ... }` → `[&]<typename emb_t, typename cache_t>() { ... }`, with matching template-parameter arities.

## Why
- Type-safe dispatch via explicit template args on lambdas (C++20) instead of opaque macro expansion.
- The fp8 emb type is now resolved through `c10::CppTypeToScalarType<fp8_e4m3_t>`, removing the `USE_ROCM` `#ifdef` split on the emb-type case (the alias alone carries the backend-specific type).
- Net **−115 lines** (macros carried significant `USE_ROCM` duplication).

### DIFF
--- a/fbgemm_gpu/codegen/inference/embedding_forward_quantized_cpu_template.cpp
+++ b/fbgemm_gpu/codegen/inference/embedding_forward_quantized_cpu_template.cpp
@@ -244,7 +244,7 @@ Tensor int_nbit_split_embedding{{ "_nobag" if nobag else "" }}_codegen_forward_{
 
     const auto* weights_tys_acc = weights_tys.const_data_ptr<uint8_t>();
 
-    DISPATCH_OUTPUT_TYPES(output.scalar_type(), "intn_split_embedding{{ "_nobag" if nobag else "" }}_codegen_forward_kernel", [&] {
+    fbgemm_gpu::dispatch_output_types(output.scalar_type(), "intn_split_embedding{{ "_nobag" if nobag else "" }}_codegen_forward_kernel", [&]<typename output_t>() {
         {% if weighted %}
         const float* indice_weights_acc = indice_weights.const_data_ptr<float>();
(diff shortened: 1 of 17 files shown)


r-barnes:
In `dispatch_emb_cache_output_types`, an inner generic lambda captures a template parameter (`output_t`) from an outer generic lambda. This pattern is valid C++20, but compilers older than GCC 12 / Clang 14 have bugs that cause a hard build error.

```cpp
// output_t is a template param of the outer lambda,
// captured into the inner lambda's closure — tricky for older compilers
const auto dispatch_emb_cache = [&]<typename output_t>() -> decltype(auto) {
    return dispatch_emb_cache_types(...,
        [&]<typename emb_t, typename cache_t>() -> decltype(auto) {
          return f.template operator()<emb_t, cache_t, output_t>(); // <-- here
        });
};
```

If the build matrix includes GCC < 12 or Clang < 14, this will fail at compile time (not runtime). Worth a toolchain version check.

cyyever:
@r-barnes These compilers are lower than the minimum requirement.

PR #6123, opened 2026-08-07
[ROCm] support warpSize 32 and 64 in a single TBE codegen build

jeffdaily:
Make the TBE codegen emit kernels and host dispatch tables that are correct for both wave32 and wave64 AMD archs from a single build, including a mixed `PYTORCH_ROCM_ARCH="gfx90a;gfx1100"` wheel. This is the codegen core of the warpSize port; the earlier PRs in the chain supplied the primitives it needs (`kWarpSizeHost`, `warpReduceAllSum` gating, cache associativity, and the CMake-derived `--has_wave32`/`--has_wave64` flags). The mechanical host-side `kWarpSize` -> `kWarpSizeHost()` launch-config substitutions were split into #6278 at reviewer request; this PR carries the codegen core and is stacked on #6278.

Two root causes are fixed here:
1. **Kernel-name mangling baked in the wave size.** The TBE kernel templates took `kThreadGroupSize` as a template integer, so its value entered the mangled symbol name; on a mixed-arch build the host pass and the gfx1100 device pass disagree and the linker cannot resolve the device symbol. The templates are reparameterised to take `kSubwarpDivisor` (a wave-size-free literal); the kernel body recovers `kThreadGroupSize = kWarpSize / kSubwarpDivisor` in the device pass.
2. **Host dispatch used the wrong wave size.** The `DISPATCH_*` macros emit a `_WAVE32`/`_WAVE64` pair selected at runtime via `kWarpSizeHost()`; single-wave-size builds emit only the matching table.

Reconciliation with main's HIP-optimized backward path (#5712, #6197): the `hip_mixed_d` warp kernel gets the same `kSubwarpDivisor` treatment as the other kernels (definition, explicit instantiations, and host selection). Its `max_D <= 128` fast path re-instantiates with a 32-lane thread group, which under this parameterisation is `kSubwarpDivisor = 2` on wave64; like the matching cta_per_row override, it is emitted only for wave64 builds and guarded by a runtime `kWarpSizeHost()` check on mixed wheels, since on wave32 the default full-warp kernel already runs 32-lane groups.

Two related wave32 device-pass fixes are folded in (num_packed_bags cap in nbit inference; runtime `max_vecs_per_thread` in the backward template for dims >= 128).

Fifth and final in the chain splitting pytorch/FBGEMM#5804 into reviewable pieces; stacked on #6278.


jeffdaily:
@q10 Friendly ping on this one -- it's been imported (D115263090) for about a week and is green/mergeable. This is the last of the five-PR split of #5804; it's ready to land whenever the internal diff can go through. Thanks!

q10:
> @q10 Friendly ping on this one -- it's been imported (D115263090) for about a week and is green/mergeable. This is the last of the five-PR split of #5804; it's ready to land whenever the internal diff can go through. Thanks!
> 
> Authored with assistance from Claude (Anthropic).

@jeffdaily yes, we are looking into this

fbgemm_gpu/codegen/inference/embedding_forward_quantized_split_lookup.cu
```diff
@@ -201,10 +201,10 @@ Tensor pruned_hashmap_lookup_cuda(
                       hash_t>),
                   utils::cuda::cap_grid_dim_x(
                       nbit::div_round_up(
-                          B * T + 1, kForwardMaxThreads / kWarpSize),
+                          B * T + 1, kForwardMaxThreads / kWarpSizeHost()),
```
q10, line 204: Can we split the change from kWarpSize to `kWarpSizeHost()` into a separate PR, in order to reduce the impact of each PR and make rollback easier if there is ever a need for rollback?

jeffdaily: Done. The kWarpSize -> kWarpSizeHost() host-launch substitutions (including the hunk on this line) are split out into #6278, and this PR is rebased on top of it so it now carries only the codegen/template reparameterisation. While rebasing over main I also reconciled with the hip_mixed_d warp kernel from #5712; the updated PR description has the details.

q10:
@jeffdaily could you rebase this PR on top of latest main to address the merge conflicts?

jeffdaily:
@q10 rebased onto latest main. The conflicts are gone and this is now a single commit, since the two commits below it landed with #6278.

The rebase needed one reconciliation worth a look: #6286 added a ROCm half-wave-group fast path to the forward dispatch macro using the old `kThreadGroupSize` parameterization, which this PR replaces with `kSubwarpDivisor`. It is now in the `_WAVE64` macro as `kSubwarpDivisor = 2`, with its explicit instantiation gated on `has_wave64`. Wave32 does not take the override because its default `MAX_D <= 128` dispatch is already a 32-lane group.

Checked by rendering the codegen for CUDA, wave64, wave32 and mixed, and comparing kernel instantiation multisets against main per file and per kernel with main's group size G mapped to this PR's divisor D via `G = warpSize / D`. CUDA and wave64 match one for one: 77644 and 90325 instantiations on both sides.

CI here is in `action_required` as well.

PR #6140, opened 2026-08-11
Avoid ATen reduction in masked jagged select (#6140)

ak7:
Summary:
X-link: https://github.com/facebookresearch/FBGEMM/pull/3041

Replace `mask.sum().item<int32_t>()` with `std::count` over the contiguous boolean mask, avoiding generic ATen dispatch, bool-to-int64 conversion, TensorIterator reduction, and temporary tensor allocation while preserving exact output sizing.

CPU benchmarks show an 8.7%-60.2% whole-operator reduction depending on mask size. 

Current cost of the whole function in the fleet is $2.456M/year

Reviewed By: bottler

Differential Revision: D115456015


bottler (approved):

PR #6160, opened 2026-08-14
Apply the last k scale block in _kernel_matmul_fp8_block_fastacc

### LINKED ISSUE #6159: matmul_fp8_block: trailing k scale block dropped for many K, results off by a per-block scale
## Summary

`_kernel_matmul_fp8_block_fastacc` applies its block scales only under `((pid_k + 1) % k_multiple == 0) or (k_remaining < BLOCK_K * SPLIT_K)`. On the final k iteration `k_remaining` equals `BLOCK_K * SPLIT_K` exactly whenever `K` is a multiple of `BLOCK_K * SPLIT_K`, so the strict `<` is false. If `K` is also not a multiple of `scale_block_k` the modulo clause is false too, and the trailing scale block is never applied — the output is off by a multiplicative per-block scale, a factor of ~1e4 with fp8 scales rather than a rounding difference.

https://github.com/pytorch/FBGEMM/blob/bef38467f88437baddad319c1c4ecba820c97e13/fbgemm_gpu/experimental/gemm/triton_gemm/fp8_gemm.py#L1853-L1854

## Reproduction

```python
import torch
from fbgemm_gpu.experimental.gemm.triton_gemm.fp8_gemm import (
    matmul_fp8_block, quantize_fp8_block,
)

M = N = 256

def run(K, block, fast_accum):
    a = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
    b = torch.randn(N, K, dtype=torch.bfloat16, device="cuda")
    aq, a_scale = quantize_fp8_block(a, block, block)
    bq, b_scale = quantize_fp8_block(b, block, block)
    out = matmul_fp8_block(aq, bq, a_scale, b_scale, block, block, block,
                           fp8_fast_accum=fast_accum)
    ref = a.float() @ b.float().T
    return ((out.float() - ref).norm() / ref.norm()).item()

for K in [256, 512, 300, 640, 896, 1152]:
    print(f"{K:>6} {K % 256:>4} {run(K, 256, True):>14.4f} {run(K, 256, False):>10.4f}")
```

```
     K   K % 256   fast_accum=True   fast_accum=False
   256         0            0.0377             0.0373
   512         0            0.0376             0.0373
   300        44            0.0375             0.0375
   640       128        10411.0859             0.0375
   896       128        11432.3818             0.0375
  1152       128        11753.1621             0.0376
```

Relative error against `a @ b.T`; ~0.04 is fp8 quantization noise. `K=300` also leaves a partial trailing scale block but is correct, because 300 is not a multiple of any candidate tile `BLOCK_K` and the `k_remaining` clause still fires. The same pattern appears at the 128-blocks `FP8TritonBlockwiseGemm` uses: `K` = 64 / 192 / 320 give 12574, 13014, 13258, while `K` = 128 / 256 are clean.

## Environment

- pytorch/FBGEMM at `bef38467f88437baddad319c1c4ecba820c97e13` (main)
- NVIDIA B200 (sm_100), driver 595.71.05
- torch 2.13.0+cu130, triton 3.7.1

truong-v:
## Description

Fixes #6159

The scale-application guard in `_kernel_matmul_fp8_block_fastacc` escapes with `(k_remaining < BLOCK_K * SPLIT_K)`, which is false on the final k iteration whenever `K` is an exact multiple of `BLOCK_K * SPLIT_K`. When `K` is also not a multiple of `scale_block_k`, the modulo clause is false as well and the trailing scale block is never applied, so the output is off by a multiplicative per-block scale.

## Verification

Relative error of `matmul_fp8_block(..., fp8_fast_accum=True)` against `a @ b.T`, M = N = 256 (~0.04 is fp8 quantization noise):

| scale blocks | K | before | after |
|---|---|---|---|
| 256 | 640 | 10411.09 | 0.0372 |
| 256 | 896 | 11432.38 | 0.0377 |
| 256 | 1152 | 11753.16 | 0.0375 |
| 128 | 64 | 12574.35 | 0.0372 |
| 128 | 192 | 13013.62 | 0.0378 |
| 128 | 320 | 13258.18 | 0.0376 |
| 256 | 256, 512, 300 | 0.0375 | 0.0375 |
| 128 | 128, 256 | 0.0375 | 0.0375 |

## Environment

- NVIDIA B200 (sm_100), driver 595.71.05
- torch 2.13.0+cu130, triton 3.7.1

### DIFF
--- a/fbgemm_gpu/experimental/gemm/test/fp8_gemm_test.py
+++ b/fbgemm_gpu/experimental/gemm/test/fp8_gemm_test.py
@@ -741,3 +741,9 @@ class TestFp8Matmul(unittest.TestCase):
         _test_matmul_fp8_block((3, 4, 5), (256, 256, 256), False)
         _test_matmul_fp8_block((3, 4, 5), (256, 256, 256), True, device="cpu")
         _test_matmul_fp8_block((1024, 2048, 4096), (256, 512, 1024), True, device="cpu")
+        # K values that leave a partial trailing scale block while still being a
+        # multiple of the tile BLOCK_K. The K=5 shapes above are not a multiple of
+        # any BLOCK_K, so they exercise a different path through the scale guard.
+        _test_matmul_fp8_block((256, 256, 640), (256, 256, 256), True)
+        _test_matmul_fp8_block((256, 256, 64), (128, 128, 128), True)
+        _test_matmul_fp8_block((256, 256, 192), (128, 128, 128), True)
--- a/fbgemm_gpu/experimental/gemm/triton_gemm/fp8_gemm.py
+++ b/fbgemm_gpu/experimental/gemm/triton_gemm/fp8_gemm.py
@@ -1851,7 +1851,7 @@ def _kernel_matmul_fp8_block_fastacc(
         # And have s_k+1 be 1.
         # Scale_i = pid_i * BLOCK_I / scale_block_i
         pid_k = k * SPLIT_K + pid_z
-        if ((pid_k + 1) % k_multiple == 0) or (k_remaining < BLOCK_K * SPLIT_K):
+        if ((pid_k + 1) % k_multiple == 0) or (k + 1 == tl.cdiv(K, BLOCK_K * SPLIT_K)):
             # Note: Due to split_k access "pid_k" = k * SPLIT_K + pid_z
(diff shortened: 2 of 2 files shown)


q10:
Landed in https://github.com/pytorch/FBGEMM/commit/44c5e744fceb7aa3f52c3f7a075e023d19f23a55

PR #6189, opened 2026-08-20
Observe shutdown on teardown instead of running loops to completion (#6189)

FriedCosey:
Summary:

X-link: https://github.com/facebookresearch/FBGEMM/pull/3078

Before this stack, base set `stop_ = true` before joining its raw threads at shutdown; the executor rewrite dropped `stop_`, so both teardown loops keep running work instead of bailing out. This diff restores parity.

```
 ~RawEmbeddingStreamer() {
+  stop_ = true;                 // base did this; the rewrite dropped it
   ...
-  consumer_executor_->join();   // drains the backlog -- N blocking
+  consumer_executor_->stop();   // co_setEmbeddings, uncapped, no timeout
 }

 while (*ptr == 0) {             // poll_copy_done_flag
   std::this_thread::yield();
+  if (stop_) return false;      // else spins the full 10s at shutdown
```

`stop()` and `join()` both enqueue poison at `LO_PRI` and join the threads; they differ only in `shouldStopThread()`, where `join()` forbids early exit so a worker must drain down to its poison task. `stop()` lets workers exit at any dequeue point -- best-effort per folly's contract ("no guarantee that unexecuted tasks won't be executed"), and in-flight ships finish either way. Producer joins stay `join()`: draining dispatch and copy is what keeps anything from scheduling onto the consumer after it stops.

Reviewed By: bottler

Differential Revision: D116397924

### DIFF
--- a/fbgemm_gpu/include/fbgemm_gpu/split_embeddings_cache/raw_embedding_streamer.h
+++ b/fbgemm_gpu/include/fbgemm_gpu/split_embeddings_cache/raw_embedding_streamer.h
@@ -11,6 +11,7 @@
 #ifdef FBGEMM_FBCODE
 #include <folly/coro/Task.h>
 #include <folly/futures/Future.h>
+#include <atomic>
 #endif
 
 #include <utility>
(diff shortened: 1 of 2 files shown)


bottler (approved):

PR #6194, opened 2026-08-21
Use distinct float16 and bfloat16 CPU template types

cyyever:
## Summary

Build on the already-upstreamed float16/bfloat16 distinct-type change (#6049) by propagating that type distinction through the embedding lookup pipeline (`EmbeddingSpMDM`/`EmbeddingSpMDMNBit`/`RefImplementations` and their autovec kernels), replacing the runtime `is_bf16_out`/`is_bf16_in` bool flags with compile-time dispatch on the distinct types.

In `src/EmbeddingSpMDMAutovec.cc`, the block-size specialization dispatch is fully switched over to upstream's newer fold-expression-based helpers (`make_spmdm_generic`/`try_spmdm_fixed_block_size`, `make_nbit_generic`/`try_nbit_fixed_block_size`, and `make_nbit_rowwise_sparse_generic`), adapted to drop the is_bf16 runtime flags those helpers still carried, replacing this branch's own macro-based SPECIALIZE/SPECIALIZE_BLOCK_SIZE system and the now-unused `specialization_helper` namespace entirely.

This branch is now rebased cleanly on current `main` and consists of exactly two commits:
- `54d118315` — mark the `is_bf16_out`/`is_bf16_in` params on `GenerateEmbeddingSpMDM*` as `[[maybe_unused]]` in `include/fbgemm/FbgemmEmbedding.h` (12 lines), in preparation for callers moving off these runtime flags. No signature change.
- `589f261b3` — the actual embedding SpMDM kernel migration (458 insertions / 671 deletions across 16 files).

Both prerequisite PRs from the earlier stack have landed on `main`:
- #6237 — landed as 86b8d3823. Note: despite its commit message, only the `src/QuantUtils.cc` piece actually landed — the `FbgemmEmbedding.h` `[[maybe_unused]]` hunk it described was dropped somewhere in the internal export/review and never made it to `main`. So that piece is **not** redundant here; it's carried forward as this branch's first commit (`54d118315`).
- #6238 — landed as 235fa376f (`QuantUtils.cc` dequant helpers). Fully redundant with this branch and already dropped.

[...]

PR #6237, opened 2026-08-28
Mark GenerateEmbeddingSpMDM* is_bf16 params [[maybe_unused]]; use to_float/from_float in QuantUtils

cyyever:
Prerequisite for #6194 ("Use distinct float16 and bfloat16 CPU template types").

## Summary

This PR originally also added a `FbgemmHalfType` concept and generic `to_float<T>`/`from_float<T>` helpers, but that part has since landed on `main` via #6049 ("Make float16/bfloat16 distinct types"), so it's been dropped from this branch as redundant. It now contains two small, independent, backward-compatible pieces:

- `include/fbgemm/FbgemmEmbedding.h`: mark the `is_bf16_out`/`is_bf16_in` parameters on `GenerateEmbeddingSpMDM*` as `[[maybe_unused]]`, in preparation for callers moving to compile-time dispatch on the distinct types instead of these runtime flags. No signature change.
- `src/QuantUtils.cc`: use the `to_float`/`from_float<T>` helpers (from #6049) instead of the equivalent direct `cpu_half2float`/`cpu_bf162float`/`cpu_float2bfloat16` calls in `FloatOrHalfToFusedNBitRowwiseQuantizedSBHalfRef`, `FloatOrHalfToFused8BitRowwiseQuantizedSBFloatRef`, `FusedNBitRowwiseQuantizedSBHalfToFloatOrHalfRef`, and `Fused8BitRowwiseQuantizedSBFloatToFloatOrHalfRef`. Purely a call-site rename to the same underlying function in each case — no behavior change. The `float16` quantize (write) branches intentionally keep the explicit `cpu_float2half_rn` call rather than `from_float<float16>` (which resolves to the hardware-accelerated `cpu_float2half` when available) to avoid changing rounding behavior, and the `cpu_half2float_ref` call in the Ref-suffixed `Fused8BitRowwiseQuantizedSBFloatToFloatOrHalfRef` scale/bias read is left untouched since it's deliberately the portable, non-hardware-accelerated reference path.

#6194 currently carries the first of these two pieces on its branch and will rebase to drop it once this lands.

## Test plan

- [x] Verified existing (pre-#6194) `src/EmbeddingSpMDM.cc`, `EmbeddingSpMDMAutovec.cc`, `EmbeddingSpMDMNBit.cc` still compile unchanged against these headers (`clang++ -fsyntax-only`) — confirms this is non-breaking for current callers.
- [x] `src/QuantUtils.cc` compiles clean with `clang++ -fsyntax-only`.
- [ ] CI green

### DIFF
--- a/include/fbgemm/FbgemmEmbedding.h
+++ b/include/fbgemm/FbgemmEmbedding.h
@@ -80,8 +80,8 @@ GenerateEmbeddingSpMDM(
     int prefetch = 16,
     bool is_weight_positional = false,
     bool use_offsets = true,
-    bool is_bf16_out = false,
-    bool is_bf16_in = false);
+    [[maybe_unused]] bool is_bf16_out = false,
+    [[maybe_unused]] bool is_bf16_in = false);
 
 /**
  * @param output_stride If -1, output_stride is same as block_size.
@@ -113,8 +113,8 @@ GenerateEmbeddingSpMDMWithStrides(
     std::int64_t input_stride = -1,
     bool scale_bias_last = true,
     bool no_bag = false,
-    bool is_bf16_out = false,
-    bool is_bf16_in = false);
+    [[maybe_unused]] bool is_bf16_out = false,
+    [[maybe_unused]] bool is_bf16_in = false);
 
 /**
  * @tparam IndexType can be int32_t or int64_t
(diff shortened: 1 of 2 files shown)


cyyever:
@q10 rebased and added more changes

PR #6261, opened 2026-09-02
[ROCm] Fix GPU memory fault in batched_unary_embeddings backward

avbokovoy:
## Problem
batched_unary_embeddings_test.py::test_gpu crashed with a GPU memory access fault on ROCm. The permute sub-test passed
  offsets of length 8 (7 segments) for a T=2 module, violating the op's offsets.numel() == T*B+1 contract. Forward silently truncated
  B=3 and dropped the trailing segment; backward walked all 7 segments, deriving an out-of-range table id t=2 that indexed
  grad_weight out of bounds.

## Fix
  - Correct the test offsets to [0..8] (9 entries -> T*B+1 for B=4).
  - Add host-side TORCH_CHECK guards on the offsets length in forward (CPU/CUDA/HIP) and backward (CUDA/HIP), so malformed input fails loudly instead of faulting/corrupting grads.

### DIFF
--- a/.github/scripts/fbgemm_gpu_test.bash
+++ b/.github/scripts/fbgemm_gpu_test.bash
@@ -168,7 +168,6 @@ __configure_fbgemm_gpu_test_rocm () {
 
   # https://github.com/pytorch/FBGEMM/issues/1559
   export ignored_tests=(
-    ./batched_unary_embeddings_test.py
     ./sll/triton_sll_test.py
     ./gather_scatter/gather_scatter_test.py
     ./moe/layers_test.py  # Not a python unittest file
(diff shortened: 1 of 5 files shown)


sylvesterkaczmarek (commented):
The eager paths now reject malformed offsets, but `batched_unary_embeddings_forward_meta` still floor-divides `(offsets.sym_numel() - 1) / T` without enforcing the new `T > 0` / divisibility invariant. FakeTensor/export can therefore accept an input shape that eager now rejects and report a misleading `B`. Could the Meta implementation mirror the new validation?

avbokovoy:
> The eager paths now reject malformed offsets, but `batched_unary_embeddings_forward_meta` still floor-divides `(offsets.sym_numel() - 1) / T` without enforcing the new `T > 0` / divisibility invariant. FakeTensor/export can therefore accept an input shape that eager now rejects and report a misleading `B`. Could the Meta implementation mirror the new validation?

Sure, done in https://github.com/pytorch/FBGEMM/pull/6261/commits/df219f48f6f55d897290f1fbbf168c5cd1d82d61. Added dedicated test in https://github.com/pytorch/FBGEMM/pull/6261/commits/99737effa2bf2eb50d0f5e54e733ad535055a4aa

PR #6278, opened 2026-09-09
[ROCm] use kWarpSizeHost() in TBE host-side launch configs

jeffdaily:
Host-side TBE launch configurations computed block/grid dims from the compile-time `kWarpSize` (or a hardcoded 64 for `BT_block_size`). In a ROCm wheel that serves both wave64 (CDNA) and wave32 (RDNA) archs the host code is compiled once, so a compile-time warp size is wrong for whichever arch it does not match: on gfx1100 these launches came up with 64-wide x-dims for 32-wide warps. This replaces those host uses with the runtime `kWarpSizeHost()`, which reports the active device's warp size. On CUDA and on wave64-only ROCm builds `kWarpSizeHost()` folds to the same values as before, so this is behavior-preserving there.

Split out of #6123 at reviewer request to keep each PR small and independently revertible. The codegen/template reparameterisation that makes the kernel symbols wave-size-agnostic stays in #6123, which is now rebased on top of this PR.

### DIFF
--- a/fbgemm_gpu/codegen/inference/embedding_forward_quantized_split_lookup.cu
+++ b/fbgemm_gpu/codegen/inference/embedding_forward_quantized_split_lookup.cu
@@ -201,10 +201,10 @@ Tensor pruned_hashmap_lookup_cuda(
                       hash_t>),
                   utils::cuda::cap_grid_dim_x(
                       nbit::div_round_up(
-                          B * T + 1, kForwardMaxThreads / kWarpSize),
+                          B * T + 1, kForwardMaxThreads / kWarpSizeHost()),
                       kForwardMaxThreads,
                       at::cuda::getCurrentCUDAStream()),
-                  dim3(kWarpSize, kForwardMaxThreads / kWarpSize),
+                  dim3(kWarpSizeHost(), kForwardMaxThreads / kWarpSizeHost()),
                   0,
                   at::cuda::getCurrentCUDAStream(),
                   PTA_B(indices, index_t, 1, 32),
@@ -266,10 +266,10 @@ Tensor pruned_array_lookup_cuda(
                       remap_t>),
                   utils::cuda::cap_grid_dim_x(
                       nbit::div_round_up(
-                          offsets.size(0), kForwardMaxThreads / kWarpSize),
+                          offsets.size(0), kForwardMaxThreads / kWarpSizeHost()),
                       kForwardMaxThreads,
                       at::cuda::getCurrentCUDAStream()),
-                  dim3(kWarpSize, kForwardMaxThreads / kWarpSize),
+                  dim3(kWarpSizeHost(), kForwardMaxThreads / kWarpSizeHost()),
                   0,
                   at::cuda::getCurrentCUDAStream(),
                   PTA_B(indices, index_t, 1, 32),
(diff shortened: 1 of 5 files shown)


jeffdaily:
@q10 heads up: the CUDA builds on this PR went red shortly after the import (D119363782), so the imported diff has the breakage in it.

Root cause: this PR added `#include "fbgemm_gpu/utils/cuda_prelude.cuh"` to `embedding_backward_split_host_template.cpp` so that `kWarpSizeHost()` would be in scope. The generated `gen_embedding_backward_split_*.cpp` are host sources compiled by the host compiler, which has no declaration for that header's device intrinsics:

```
cuda_prelude.cuh:186:10: error: use of undeclared identifier '__ballot_sync'
cuda_prelude.cuh:230:3:  error: use of undeclared identifier '__syncwarp'
```

ROCm builds did not catch it because their host pass is hipcc, which does declare them. The CPU-only build variant would have failed as well, since `cuda_prelude.cuh` pulls in `<cuda.h>`.

Fix pushed as 60f547b1: `kWarpSize` and `kWarpSizeHost` move verbatim into a new `fbgemm_gpu/utils/warp_size.h` that `cuda_prelude.cuh` includes, so every device-side user is unaffected. The host template includes the new header instead, and the `kWarpSizeHost()` call site is unchanged. The new header needs no CUDA or HIP include of its own, because under `USE_ROCM` `at::cuda::warp_size()` is declared by `<c10/macros/Macros.h>`; running hipify over the moved block produces no substitutions, so leaving it as a `.h` outside the hipified set is safe.

One thing worth a check on your side: this adds a new file under `fbgemm_gpu/include/fbgemm_gpu/utils/`. OSS CMake picks it up with no changes, but I cannot see whether the fbcode target globs that directory.

Could you re-review and re-import when you get a chance? The GitHub workflows are currently sitting in `action_required` and need an approval to run.

Separately, #6123 (the codegen half of the split) is now rebased onto current main and stacked on this one, and carries the same fix, so it will need a re-import too.

q10:
Hi @jeffdaily I think we missed the last commit 60f547b you pushed to the PR when we landed the dif (we had addressed the build issues internally, but they aren't the same code changes as in 60f547b.  Could you open 60f547b as a new PR on top of latest main, if the changes are still applicable?

jeffdaily:
@q10 opened as #6323.

Mostly applicable, with one correction worth flagging. The header split on its own would have been a no-op on current main: your version removed the `kWarpSizeHost()` call site from the host template rather than fixing the include, so nothing in a host `.cpp` needs the header any more.

What is still needed is that the landed diff deleted the whole `#ifdef USE_ROCM` block, which moved ROCm from `BT_block_size` 64 and `max_segment_length_per_warp` 16384 to the CUDA values of 32 and 32. Both predate this work, and `max_segment_length_per_warp` is the long-run threshold fed to the unique-index counting kernel, so 32 sends most segments down the `cta_per_row` path on ROCm. #6323 restores the block, with `BT_block_size` as the runtime query this PR set out to use everywhere else, and carries `warp_size.h` as the piece that makes that callable from a host source.

The other four files landed as a behavior-equivalent refactor, so #6323 does not touch them.

Its workflows are sitting in `action_required` and need an approval to run.

PR #6288, opened 2026-09-10
Fix GenAI ROCm build

avbokovoy:
Wrap hipMemcpy with C10_CUDA_CHECK so `[[nodiscard]]` return value is not discarded


q10:
Merged in https://github.com/pytorch/FBGEMM/pull/6201

PR #6295, opened 2026-09-11
Add FTRL-Proximal as a fused TBE optimizer (CUDA + CPU), matching NVIDIA DynamicEmb

tiankongdeguiji:
Adds `OptimType.FTRL` (FTRL-Proximal) as a fused TBE optimizer, with both CUDA and CPU backends.

FTRL-Proximal is the standard optimizer for the sparse/wide part of CTR models, and is what Alibaba's [x-deeplearning](https://github.com/alibaba/x-deeplearning) uses for its embedding parameter server. There was no equivalent in FBGEMM, so recsys workloads that rely on it have nothing to migrate onto.

This implementation matches **NVIDIA DynamicEmb's FTRL** ([NVIDIA/recsys-examples#487](https://github.com/NVIDIA/recsys-examples/pull/487)) element for element. That PR's `construct_twin_module` currently raises `ValueError` on `"ftrl"` because FBGEMM has no counterpart, so this is what would let DynamicEmb build a TorchRec twin for FTRL models.

## Algorithm

Algorithm 1 of McMahan et al. 2013 (*Ad Click Prediction: a View from the Trenches*), with the paper's fixed square root generalized to an arbitrary exponent the way TensorFlow's `FtrlOptimizer` does. Per element, with state `linear` (z) and `accum` (n):

```
new_accum = accum + g^2
sigma_new = new_accum ^ (-learning_rate_power)
sigma_old = accum     ^ (-learning_rate_power)
linear   += g - (sigma_new - sigma_old) / learning_rate * weight
accum     = new_accum
quadratic = (ftrl_beta + sigma_new) / learning_rate + l2_reg
weight    = |linear| > l1_reg ? (l1_reg * sgn(linear) - linear) / quadratic : 0
```

`learning_rate`, `ftrl_beta`, `ftrl_l1_reg` and `ftrl_l2_reg` are the paper's alpha, beta, lambda1 and lambda2. `learning_rate_power == -0.5` (the default) recovers the paper's `alpha / (beta + sqrt(n))` exactly and takes the faster `sqrtf` path.

Note that TensorFlow and x-deeplearning scale their l2 knob differently -- their quadratic term is `2 * l2` with no beta -- so a configuration ported from either must halve its l2 value.

Two properties worth calling out, both inherent to the algorithm rather than artifacts of this implementation:

- FTRL **recomputes** the weight from `(linear, accum)` instead of incrementing it, so a row's initial random value is discarded the first time that row is touched.
- The update divides by `learning_rate`, so `learning_rate > 0` is required. The TBE constructor asserts this for FTRL rather than letting it produce inf/NaN weights.

## State layout

`momentum1` = `linear`, `momentum2` = `accum`, both **elementwise** (`D` floats per row), the same footprint as `ADAM` -- and the same order DynamicEmb uses, so the two line up positionally. `get_optimizer_state()` reports `linear` then `accum`.

Because `momentum1` holds `linear`, which must start at zero, a framework that seeds optimizer state by buffer name must not point a `momentum1` seed at FTRL.

## Knobs

Three new TBE constructor arguments, all defaulted so they are no-ops for every other optimizer:

| arg | default | meaning |
|---|---|---|
| `ftrl_learning_rate_power` | `-0.5` | exponent applied to `accum`; -0.5 gives the classic 1/sqrt schedule. Must be `<= 0` -- a positive value would make the learning rate grow without bound, and is rejected at construction |
| `ftrl_beta` | `0.0` | the paper's beta; bounds the per-coordinate learning rate while `accum` is still small |
| `ftrl_l1_reg` | `0.0` | L1 strength; a coordinate whose `\|linear\|` never exceeds this stays exactly `0`, which is what makes FTRL sparsify embedding rows |
| `ftrl_l2_reg` | `0.0` | proximal L2 strength, the paper's lambda2 (coefficient 1, not TF's `2 * l2`) |

They are `ftrl_`-prefixed because `OptimizerArgs` is a flat namespace shared by every optimizer. `weight_decay` is deliberately **not** reused for `ftrl_l2_reg`: FBGEMM's `weight_decay` is half of a `(weight_decay, weight_decay_mode)` pair with six modes and enters as a *gradient* term, whereas FTRL's L2 enters the *denominator* and honors none of those modes. Overloading it would make `weight_decay_mode` a silent no-op on FTRL tables.

`initial_accumulator_value` is intentionally not implemented — no FBGEMM optimizer has a state-init argument today (Adagrad included), so an FTRL-only one would be inconsistent. A generic `optimizer_state_init_values` TBE argument would be the right way to add it later.

## Support matrix

| | |
|---|---|
| CUDA | yes |
| CPU | yes |
| VBE | yes |
| global weight decay | no — FTRL's proximal L2 is not a gradient-space decay, so GWD does not compose with it |
| SSD / optimizer offloading | no (the `EmbOptimType` metadata arms are filled in, so enabling it later is mechanical) |

## Tests

`fbgemm_gpu/test/tbe/training/backward_optimizers_test.py`:

- `test_backward_optimizers_ftrl` — the shared randomized harness over variable T/D/B/E/L, mixed dims, weighted/unweighted, VBE and all pooling modes, on CPU and CUDA. The new reference block recomputes `accum`, `linear` and the post-step weights from the formula above and compares against `split_optimizer_states()` / `split_embedding_weights()`.
- `FtrlTest` — a deterministic class covering what the randomized harness cannot check cheaply: `ftrl_l1_reg` producing *exactly* `0.0`, the generic `powf` path agreeing with the reference (and differing from the sqrt fast path), the `ftrl_beta` and `ftrl_l2_reg` denominator terms, multi-step `linear`/`accum` accumulation, zero-gradient idempotence, rejection of a positive `learning_rate_power`, and CPU/CUDA parity.

Verified on an A10 (sm86) against a verbatim port of DynamicEmb's own reference (`ftrl_step` in recsys-examples `test/unit_tests/optimizer/test_ftrl_optimizer.py`, float64) across all five parameter sets it parametrizes over — `plain`, `beta_l2`, `l1`, `power`, `seeded_accum` — on both CUDA and CPU: **10/10 exact**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FPQnjDECjd8ZEvGF5Hpzpn

### DIFF
--- a/fbgemm_gpu/fbgemm_gpu/split_embedding_configs.py
+++ b/fbgemm_gpu/fbgemm_gpu/split_embedding_configs.py
@@ -71,6 +71,7 @@ class EmbOptimType(enum.Enum):
     PARTIAL_ROWWISE_LAMB = "partial_row_wise_lamb"
     ROWWISE_ADAGRAD = "row_wise_adagrad"
     ROWWISE_RMSPROP_AR = "row_wise_rmsprop_ar"
+    FTRL = "ftrl"
     SHAMPOO = "shampoo"  # not currently supported for sparse embedding tables
     SHAMPOO_V2 = "shampoo_v2"  # not currently supported for sparse embedding tables
     MADGRAD = "madgrad"
@@ -100,7 +101,11 @@ class EmbOptimType(enum.Enum):
             return ["momentum1"]
         elif self == EmbOptimType.ROWWISE_RMSPROP_AR:
             return ["momentum1", "prev_iter"]
-        elif self in [EmbOptimType.PARTIAL_ROWWISE_ADAM, EmbOptimType.ADAM]:
+        elif self in [
+            EmbOptimType.PARTIAL_ROWWISE_ADAM,
+            EmbOptimType.ADAM,
+            EmbOptimType.FTRL,
+        ]:
             return ["momentum1", "momentum2"]
         else:
             return []
@@ -116,7 +121,7 @@ class EmbOptimType(enum.Enum):
             return {"momentum1": 1, "prev_iter": 1}
         elif self == EmbOptimType.PARTIAL_ROWWISE_ADAM:
             return {"momentum1": D, "momentum2": 1}
-        elif self == EmbOptimType.ADAM:
+        elif self in (EmbOptimType.ADAM, EmbOptimType.FTRL):
             return {"momentum1": D, "momentum2": D}
         else:
             return {}
@@ -143,7 +148,7 @@ class EmbOptimType(enum.Enum):
         elif self == EmbOptimType.PARTIAL_ROWWISE_ADAM:
             return pad4(1 * momentum2_dtype.itemsize) + D * momentum1_dtype.itemsize
 
-        elif self == EmbOptimType.ADAM:
+        elif self in (EmbOptimType.ADAM, EmbOptimType.FTRL):
             return (D * momentum1_dtype.itemsize) + (D * momentum2_dtype.itemsize)
 
         else:
@@ -188,7 +193,7 @@ class EmbOptimType(enum.Enum):
                 ),
             }
 
-        elif self == EmbOptimType.ADAM:
+        elif self in (EmbOptimType.ADAM, EmbOptimType.FTRL):
             # momentum2 lies after momentum1
             p1 = p0 + (D * momentum1_dtype.itemsize)
 
@@ -273,6 +278,11 @@ class EmbOptimType(enum.Enum):
                 "momentum1": table_size_cumsum,
                 "momentum2": table_size_cumsum,
                 "row_counter": row_count_cumsum,
+            }
+        elif self == EmbOptimType.FTRL:
+            params = {
+                "momentum1": table_size_cumsum,
+                "momentum2": table_size_cumsum,
             }
         else:
             params = {}
--- a/fbgemm_gpu/cmake/tbe_sources.py
+++ b/fbgemm_gpu/cmake/tbe_sources.py
@@ -14,6 +14,7 @@ COMMON_OPTIMIZERS = [
     "sgd",
     "rowwise_adagrad_with_counter",
     "rowwise_rmsprop_ar",
+    "ftrl",
 ]
 
 # To be populated in the subsequent diffs
@@ -65,6 +66,7 @@ VBE_OPTIMIZERS = [
     "dense",
     "adam",
     "rowwise_rmsprop_ar",
+    "ftrl",
 ]
 
 # Individual optimizers (not fused with SplitTBE backward)
--- a/fbgemm_gpu/codegen/genscript/generate_backward_split.py
+++ b/fbgemm_gpu/codegen/genscript/generate_backward_split.py
@@ -404,6 +404,7 @@ class BackwardSplitGenerator:
             rowwise_adagrad_with_counter(),
             approx_rowwise_adagrad_with_counter(),
             rowwise_rmsprop_ar(),
+            ftrl(),
             rowwise_weighted_adagrad(),
             sgd(),
             approx_sgd(),
--- a/fbgemm_gpu/codegen/genscript/optimizers.py
+++ b/fbgemm_gpu/codegen/genscript/optimizers.py
@@ -962,6 +962,178 @@ def rowwise_rmsprop_ar() -> dict[str, Any]:
     }
 
 
+def ftrl() -> dict[str, Any]:
+    """
+    FTRL-Proximal (Follow The Regularized Leader).
+
+    Per-element update, identical to TensorFlow's `ApplyFtrlV2` (with
+    `l2_shrinkage = 0`) and to Alibaba x-deeplearning's `FtrlUpdater`:
+
+        new_accum = accum + g^2
+        sigma_new = new_accum ^ (-learning_rate_power)
+        sigma_old = accum     ^ (-learning_rate_power)
+        linear   += g - (sigma_new - sigma_old) / learning_rate * weight
+        quadratic = sigma_new / learning_rate + 2 * l2_reg
+        weight    = |linear| > l1_reg ? (l1_reg * sgn(linear) - linear) / quadratic : 0
+        accum     = new_accum
+
+    `learning_rate_power == -0.5` (the default) is special-cased to `sqrt`,
+    which is faster and is what x-deeplearning does.
+
+    Per-element state: momentum1 (`accum`, the running sum of squared
+    gradients) and momentum2 (`linear`). 2 * D floats per row, the same
+    footprint as ADAM.
+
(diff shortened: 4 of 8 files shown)


tiankongdeguiji:
The TorchRec side that consumes this is meta-pytorch/torchrec#4710 — it adds a dense `torchrec.optim.FTRL` counterpart, the optimizer-class ↔ `EmbOptimType` mapping, and the planner's optimizer-state multiplier. That PR gates its registration on `hasattr(EmbOptimType, "FTRL")` so it stays importable against fbgemm_gpu wheels that predate this change.

tiankongdeguiji:
hi, @q10 @TroyGarden @aliafzal , could you take a look?

fbgemm_gpu/fbgemm_gpu/split_embedding_configs.py
```diff
@@ -274,6 +279,11 @@ def ssd_state_splits(
                 "momentum2": table_size_cumsum,
                 "row_counter": row_count_cumsum,
             }
+        elif self == EmbOptimType.FTRL:
+            params = {
+                "momentum1": table_size_cumsum,
+                "momentum2": table_size_cumsum,
+            }
```
TroyGarden, line 286: it seem the "self == EmbOptimType.FTRL" condition will never hit?

q10:
Hi @tiankongdeguiji thanks for making this contribution.  Adding a new optimizer is a large change, and after some deliberation, we believe it is best to hold off on this PR for now.  Adding a a new optimizer has been the last option for us for the following reasons:

1. Build size - A new optimizer results in a new code pipeline, and bc FBGEMM is largely compiled, this will expand the binary size.  We have hit multiple symbol relocation and other issues related to large binary size in the past, and have had to remove support for non-common optimizers as a result.
1. Maintenance - If added, we will have additional maintenance burden for an optimizer that we might not end up using, and that we believe that haphazard code maintenance would be a disservice to the community

We recommend two alternate solutions in the meantime:

1. Reuse and extend an existing optimizer that is similar enough to support FTRL-Proximal (though it looks like the algorithm doesnt closely match any existing optimizer).
1. Build a custom package with the new optimizer added.  We have comprehensive docs on building FBGEMM at https://docs.pytorch.org/FBGEMM/fbgemm_gpu/development/BuildInstructions.html, and a one-invocation build script available in https://github.com/pytorch/FBGEMM/blob/main/ci/integration/fbgemm_gpu_oss_build.bash for this.

Let us know if you have any questions, and thank you for your understanding.

tiankongdeguiji:
> Hi @tiankongdeguiji thanks for making this contribution. Adding a new optimizer is a large change, and after some deliberation, we believe it is best to hold off on this PR for now. Adding a a new optimizer has been the last option for us for the following reasons:
> 
> 1. Build size - A new optimizer results in a new code pipeline, and bc FBGEMM is largely compiled, this will expand the binary size.  We have hit multiple symbol relocation and other issues related to large binary size in the past, and have had to remove support for non-common optimizers as a result.
> 2. Maintenance - If added, we will have additional maintenance burden for an optimizer that we might not end up using, and that we believe that haphazard code maintenance would be a disservice to the community
> 
> We recommend two alternate solutions in the meantime:
> 
> 1. Reuse and extend an existing optimizer that is similar enough to support FTRL-Proximal (though it looks like the algorithm doesnt closely match any existing optimizer).
> 2. Build a custom package with the new optimizer added.  We have comprehensive docs on building FBGEMM at https://docs.pytorch.org/FBGEMM/fbgemm_gpu/development/BuildInstructions.html, and a one-invocation build script available in https://github.com/pytorch/FBGEMM/blob/main/ci/integration/fbgemm_gpu_oss_build.bash for this.
> 
> Let us know if you have any questions, and thank you for your understanding.

@q10 @TroyGarden 
Thanks for the detailed explanation. We understand the concerns around both binary size and the long-term maintenance cost of adding another fused optimizer.

I wanted to provide a bit more context on why we are proposing FTRL upstream rather than maintaining it purely as a custom FBGEMM package.

We are the Alibaba Cloud recommendation systems team and have been maintaining [[TorchEasyRec](https://github.com/alibaba/TorchEasyRec)](https://github.com/alibaba/TorchEasyRec) as our PyTorch-based recommendation stack. A significant portion of our advertising/recommendation users are migrating from Alibaba XDL to the PyTorch/TorchRec ecosystem. For these users, FTRL-Proximal is not just an additional optimizer option: it is part of the existing production training semantics, especially for sparse/wide components, so preserving FTRL behavior is important for migration compatibility.

More broadly, we believe native FTRL support in FBGEMM would also be useful to the advertising and recommendation community in the PyTorch ecosystem, beyond Alibaba Cloud users. FTRL has been widely used for sparse features and wide components in large-scale advertising systems, so having it available through the standard FBGEMM/TorchRec stack would make it easier for existing production workloads to migrate to and adopt the PyTorch ecosystem.

A custom FBGEMM package is technically possible for us, but the long-term cost is more than maintaining a single kernel. Supporting FTRL also requires corresponding integration on the TorchRec side, as in https://github.com/meta-pytorch/torchrec/pull/4710, including the optimizer mapping, optimizer-state accounting in the planner, and fused embedding integration. Maintaining downstream forks of both FBGEMM and TorchRec across releases would introduce a fairly large compatibility and maintenance surface for us and for users migrating to the upstream Torch ecosystem.

Regarding the maintenance concern: our team is willing to actively maintain the FTRL-specific code paths. If future FBGEMM changes break FTRL, or if there are FTRL-related bugs or CI failures, we are happy to investigate and submit fixes. You can also reach us at [[hongsheng.jhs@alibaba-inc.com](mailto:hongsheng.jhs@alibaba-inc.com)](mailto:hongsheng.jhs@alibaba-inc.com) if coordination is needed.

We completely understand the binary-size concern, especially given that adding a fused optimizer generates a number of specialized kernels. If binary size is the main blocker, would an **opt-in build approach** be acceptable?

For example, FTRL could live upstream but be excluded from the default FBGEMM build/wheels, with something along the lines of an optional optimizer/build flag. The default build would therefore pay no binary-size cost, while distributions that need FTRL, such as our Alibaba Cloud builds, could explicitly enable it. We would be happy to implement the build gating as part of this PR.

Would an upstream-but-opt-in implementation along these lines address the build-size and maintenance concerns?

fbgemm_gpu/fbgemm_gpu/split_embedding_configs.py
tiankongdeguiji: Good catch. You're right that this branch is not currently reachable, since FTRL has has_ssd_support=False in this PR.

I added it for symmetry with ROWWISE_RMSPROP_AR, which also has has_ssd_support=False but still defines its state layout in ssd_state_splits(). The intent was to keep the optimizer-state metadata complete and make future SSD support mechanical.

Happy to remove it if you prefer this helper to only cover currently supported SSD optimizers.

PR #6319, opened 2026-09-17
Refresh the release runner groups on release tags

huydhn:
The OSDC release runner groups are org-level and allow-list workflows by ref, so
a repo building wheels on release runners has to keep its own entries current or
its release builds queue forever waiting for a runner that will never accept
them. This runs the shared reconciler on every release tag, and on demand.

--repo scopes the run to this repository's entries: the groups are shared with
pytorch/pytorch and the other repos already on OSDC, and a run here must not
touch theirs.

Dry-run by default on workflow_dispatch and on pull requests that touch this
file, so the discovery output can be read before anything is applied. Tag
pushes apply.

Depends on pytorch/test-infra#8826. FBGEMM has no release/X.Y branch to anchor
on -- its only release/ branch is release/pytorch-1.5 -- so until that lands the
reconciler exits before doing anything.

Authored with Claude Code.


huydhn:
Not needed. The release runner groups restrict on the workflow that actually runs the job — the reusable one — not the caller. FBGEMM's wheel pipelines all call `pytorch/test-infra/.github/workflows/build_wheels_linux.yml@main`, which is already in the allowlist, so FBGEMM needs no entries of its own:

```
meta-prod-aws-ue1-release-runners: 0 FBGEMM entries of 77, yet the jobs run
pytorch/test-infra/.github/workflows/build_wheels_linux.yml@refs/heads/main   <- admitted through this
```

That holds for release tags too: the entry pins the reusable workflow's ref, not the caller's, so a `v1.2.3` build resolves to the same line. There is nothing per-release to reconcile.

